Issue #6, opened 2013-10-04
Fix display of newly added API key.


amcgregor:
Currently (in development mode) displays null's.

PR #25, opened 2013-10-12
Bugfix for #6 - Values as 'undefined' for key and vcode after adding new API key

mglinski:
It's fixed, but not for long!


amcgregor:
Tabs or spaces, don't mix that shit, yo!  Otherwise looks great.

amcgregor:
Pulled and dev server updated.

Issue #7, opened 2013-10-04
Automatically import characters when adding a key.


amcgregor:

Byetsniv:
Weehee, new version of my patch.

PR #30, opened 2013-10-20
Issue #7 : Add characters when adding keys

Byetsniv:
Added importChars() function to EVECredential, which imports all characters
attached to that key.

Whenever a key is added via the GUI, importChars() is called.

Also added "get_or_create" type functions to EVEAlliance and EVECorporation.

My first commit, so please point out things that should be different (coding style or anything).  Also, this is very much back-end. There is a "characters/list.html" frontend but that's for testing so anyone can fix it up.

Also, not sure what details were wanted here. I only store name, corp and alliance. If something else should be stored, let me know.

### DIFF
--- a/adam/auth/controller/character.py
+++ b/adam/auth/controller/character.py
@@ -8,10 +8,9 @@ from marrow.util.bunch import Bunch
 
 from adam.auth.model.eve import EVECredential
 from adam.auth.util.predicate import authorize, authenticated, is_administrator
-from adam.api import *
 
 
-class KeyInterface(HTTPMethod):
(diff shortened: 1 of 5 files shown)


amcgregor:
Could you merge current master (heh, almost said trunk) and re-issue the pull request?  I'm, apparently, lazy and just want to push a shiny green button on GitHub.  Otherwise, looks good barring a minor code style issue that is completely irrelevant.

Byetsniv:
Master merged, feel free to pull. 

Also, feel free to point out code style issues, I'd rather get used to doing things right than to have you fixing all my stuff.

PR #71, opened 2014-02-04
Provide more detailed instructions in the Add Key modal

ghost:
Additionally, explain how Core uses API keys on the /key page


amcgregor:
@kiu This isn't the right ticket for that comment, and OpenSSL generates PEM-encoded keys which are unfortunately not wholly acceptable.  Re-opening #76.

PR #81, opened 2014-02-06
Added password length check for 'password change' in settings for consistency.

Funi1234:
Quick update for consistency.
On sign up users are limited to a password of 100 chars.

Changes made to keep this limitation when users go to change their password from the settings screen.


amcgregor:
Validation of length is a purely client-side concern in this pull request; server-side validation should be added in the future.

PR #82, opened 2014-02-06
Key page now adds new keys to display list at the moment of submission + Serverside Password check 

Funi1234:
This only works with a correct Eve key. 
A 500 server error is returned if key is not valid and entry not added to list on submission.

The invalid entry is however added upon refresh of the page.


amcgregor:
Next step would be to capture a bad EVE key and present a message to the user instead of 500'ing, but his is looking great!

PR #86, opened 2014-02-14
Fix exception & make refresh inline when adding keys

noilek:
When adding a key an exception due to dereferencing a key not in the
bunch is thrown, this has been fixed.

Also make refreshing unnecessary by pulling API details inline.


amcgregor:
The .pull() call should happen asynchronously in the future, thus this fix won't find any characters.  However for now this should be a nice addition.

PR #87, opened 2014-02-20
updated readme

rdurham1007:
added pip command to add the relaxml package


rdurham1007:
simple change, just making sure I understood the pull request process.

amcgregor:
Yup, this pull request could be merged.  However the problem it solves has already been corrected by replacing the old XML parser dependency with the new one in `setup.py`.  See: https://github.com/bravecollective/core/blob/develop/setup.py#L32

Issue #57, opened 2014-02-02
Implement password recovery.


kiu:
1. Go to https://core.bravecollective.net/account/
2. Click Password Recovery

Expected
Some kind of password recovery functionality. E.g. email address input field.

Actual
Nothing happens

Comment
Due to the fact that https://github.com/bravecollective/core/issues/53
it is even more important to be able to recover your password.

ghost:
+1

jboning:
This issue can be closed now, right? :D

Acidity:
Tested by Khartana, and live on SOUND's Core deployment. Resolved by PR #98.

Issue #40, opened 2013-12-20
User names must be case-insensitive.


ghost:
If you login with a username/password at [`/account/authenticate`](http://localhost:8080/account/authenticate) the capitalization of the username must match that on the account. The username should match case insensitively, and should be automatically lower-cased at all entry points for user name data.

Acidity:
I'll have a go at resolving this.

jboning:
What's the motivation for case insensitivity?

Acidity:
Usernames and email addresses are typically case insensitive, and we were having a lot of support issues where people would forget the exact capitalization of the username that they signed up.

jboning:
Ugh. Case insensitivity is an abomination. I guess if it lightens the support load, then it's probably a win, though. And it seems that everyone else on the internet does it.

Did you know that case transformation behaviour changes between different unicode versions? Have fun upgrading python!!

ghost:
eve character names are also case insensitive

there's not much, if any? reason for a case similar username

jboning:
Yeah, it's probably fine. I've just seen firsthand (at work) some terrible consequences of case-insensitivity so I'm a bit allergic to the idea of using case insensitivity anywhere.

jboning:
I just deployed SOUND's instance up to the latest on develop, and had problems with backwards incompatibility due to the case sensitivity changes. (I haven't 100% confirmed that this was the issue, but it fits my observations.) This requires a database migration to convert all existing users to have lowercase.

Acidity:
That is correct. This patch does require the database administrator change all existing email and usernames in the database to lowercase (and resolve any potential conflicts that might arise in doing so).

Tomok:
The same will apply for the fix for #57 since that assumes lowercase e-mails. Sorry for not mentioning that in a commit message.

amcgregor:
Canonicalization of potentially unicode data is, certainly, an abomination unto the eyes of the FSM.  And yes, we had a user sign up with the name "nyao~~~~🐱" (emoji cat face), for real-real.

That is, however, one support case vs. the many others where users are too silly to be able to remember the case of their username… or the case of their e-mail address.

Besides, canonicalization of e-mail address components basically means lower-case anyway (at least for the domain part and the user part excluding + note.)  It's a tough problem.

ghost:
as OP, i feel I should contribute!
i don't like coercing existing usernames and only storing lowercase usernames.

[...]

PR #98, opened 2014-03-18
Password recovery (Issue #57)

Tomok:
Simple password recovery functionality using e-mail.
Might require some logging - just let me know.

Also please review the JavaScript - I'm not very familiar with that language.

### DIFF
--- a/brave/core/account/authentication.py
+++ b/brave/core/account/authentication.py
@@ -122,7 +122,7 @@ def send_recover_email(user):
     # send email - TODO
     params = urllib.urlencode({'email': user.email,
                                'recovery_key': str(recovery_key)})
-    url = format(u"http://127.0.0.1:8080/account/recover_password?" +
+    url = format(u"http://127.0.0.1:8080/account/recover?" +
                  str(params))
     log.info(u"RecoverURL= "+url)
 
--- a/brave/core/account/authentication.py
+++ b/brave/core/account/authentication.py
@@ -10,13 +10,15 @@ Yubikey OTP token and (regardless of identifier) a password.  Actual password cr
 
 from __future__ import unicode_literals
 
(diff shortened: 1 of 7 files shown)


amcgregor:
Awesome; I'll have a gander at this first thing after work.  Were you able to test the message content locally before pushing?

Tomok:
Yes, I checked both maildir and smtp, though I had to use the
parameter-part of the url only and paste it to my localhost domain.
Though I'm pretty sure that will not cause any problems.

Am 2014-03-18 19:02, schrieb Alice Zoë Bevan–McGregor:

amcgregor:
Just a heads up that https://github.com/bravecollective/core/blob/develop/brave/core/util/signal.py#L42-L43 configures and starts the mailer interface automatically once on application startup.  (So no need to import the Mailer class itself.)  Marrow Mailer is also [intelligent enough to apply configuration-derived message defaults if you use the resulting mailer interface instance to construct new messages](https://github.com/marrow/marrow.mailer/blob/develop/marrow/mailer/__init__.py#L150), so it unfortunately eliminates the need for all of your configuration mangling code.  For details, see the section right above this one:

https://github.com/marrow/marrow.mailer/#422-read-only-attributes

I'll continue to review as I get a chance through the day.  :)

Tomok:
Thanks for the pointers, implemented those changes.

Tomok:
still any points open, that I should resolve for this pull request ?

Acidity:
Make sure that user provided usernames and email addresses are lowercase pursuant to issue #40

Tomok:
merged current brave/develop & lowercased e-mail (recovery does not do any lookups using the username)
Also added a simple regex to e-mail field to ensure that e-mails are lowercase

PR #112, opened 2014-04-16
Adds minimum password entropy requirements to all new user accounts.

Acidity:
Also introduces new validation.js file, and updates the signup form to utilize that.


Acidity:
New password entropy requirements do not apply to old user accounts, and the sign in page remains unaffected.

PR #122, opened 2014-04-23
Add zxcvbn as python dependency

jboning:


jboning:
abandoning since I rolled this into #125

PR #127, opened 2014-04-25
Fixed email addresses.

kiu:
- Replaced author with noreply@braveineve.com
- Replaced bcc with commented out noreply@example.com
- Replaced blocked user with hostmaster@braveineve.com

### DIFF
--- a/brave/core/api/core.py
+++ b/brave/core/api/core.py
@@ -101,7 +101,7 @@ class CoreAPI(SignedController):
             return dict(
                     status = 'error',
                     code = 'blacklist',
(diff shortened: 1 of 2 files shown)


brave/core/api/core.py
```diff
@@ -101,7 +101,7 @@ def authorize(self, success=None, failure=None):
             return dict(
                     status = 'error',
                     code = 'blacklist',
-                    message = "You have been blacklisted.  To dispute, contact hostmaster@bravecollective.net"
+                    message = "You have been blacklisted.  To dispute, contact hostmaster@braveineve.com"
```
jboning, line 104: can we put this in the config file instead?

Issue #12, opened 2013-10-04
ACL CRUD management.


amcgregor:
- [ ] Basic ACL CRUD views.
- [ ] Basic ACL CRUD actions.

Byetsniv:
I'll try to get to work on the CRUD actions with some dirty views tonight (my time). Current thinking on ACLs is as follows, someone say something if you have better ideas, this is just my third draft of this:

A Permission object is a python dict() (hash table if you don't know python) as follows

``` python
{
  'type': 'user',   # or 'group' 
  'name': 'Byetsniv',
  'allowed': View # or Edit/Full or something else? More? we can define these constants somewhere.
}
```

Each "view"/page/feature will have an ACL, which is just a python list() of Permission objects.

Checking an ACL will simply be a matter of going through the list as follows:

``` python
def matches(user, perm):
  if perm['type'] == 'user':
    return perm['name'] == user.name
  if perm['type'] == 'group':
    return user.groups.contains(perm['name'])
  return False

def get_permission(current_user, ACL)
  for perm in ACL:
    if matches(current_user, group):
      return perm['allowed']
  return False
```

ACL operations then (in python) become list operations, relatively straight forward to manage. 

# Downsides:

Managing all ACLs might be cumbersome in terms of complexity. Looking up to see what Byetsniv is allowed to do will involve checking the ACL on each view/page. I don't think this is a real problem, as we shouldn't have too many views anyway, and it's hopefully not something people will do often.

# Questions I've considered:

## Should I write code so that permissions might be one of "None", "View", "Full"?

The two options I see are:
- Byetsniv has permission to access "ViewTopSecretThing", but does not have permission to access "EditTopSecretThing", or
- Byetsniv has the permission level "ViewOnly" to "TopSecretThing".

I've gone with the second. Consider the case where a user is loading a page of Inventory. If they have ViewOnly they should get a read-only page, whereas if they have Full permission they might be shown extra buttons. If I'd gone with the first option, I'd have to check for "ViewInventory" permission and "EditInventory" permission.

## Should we use constant variables rather than strings for comparators?

My code compares the strings 'user' and 'group' often.  I've noted that I'll make View and Full actual constants, but I'm not sold either way. Comparing constants should be faster. Maybe python enums? Personally I'm leaning towards constants or enums, but I'm a C++ coder so I'm not 100% how well we can do that in Python. Google has some solutions though.

Anyway, comments welcome, T-4 hours till I get anywhere near coding anything.

[...]

PR #133, opened 2014-04-27
The world's worst group and acl UI

jboning:
This is a bare minimum. And I'm not terribly confident in the code (I
haven't double-checked everything carefully yet). But, I think it's
enough to get things going!


jboning:
kiiiind of addresses #12.

Acidity:
Bug: Adding a character and then removing that character leads to a Rules: set of "grant if character is" instead of a blank rules set as found in newly created groups.

Acidity:
Just noticed this, but there does not appear to be a way to remove groups in the interface.

PR #134, opened 2014-04-30
Overrode delete function for the User class to now delete the User's...

Acidity:
...associated data such as characters and API keys.


Acidity:
Closed because I'm a moron

PR #135, opened 2014-04-30
Overrode delete function for the User class to now delete the User's associated data such as characters and API keys.

Acidity:


brave/core/account/model.py
```diff
@@ -110,6 +110,14 @@ def merge(self, other):
         
         other.delete()
 
+    def delete(self):
```
jboning, line 113: I think this change may actually not be necessary: http://docs.mongoengine.org/apireference.html#mongoengine.fields.ReferenceField Should confirm that the reverse_delete rule is set properly on all these other schemas, though.

Acidity, line 113: I know from testing that deleting the User object does not delete Characters and Keys associated with that object, so I'm going to merge this for now and we can revisit this topic later (plus, explicit is better than implicit).

PR #136, opened 2014-04-30
Added merge account button to the account settings page.

Acidity:

### DIFF
--- a/brave/core/account/controller.py
+++ b/brave/core/account/controller.py
@@ -335,6 +335,42 @@ class Settings(HTTPMethod):
                 return 'json:', dict(success=False, message=_("Invalid email address provided."), data=data)
             except NotUniqueError:
                 return 'json:', dict(success=False, message=_("The email address provided is already taken."), data=data)
+        
+        #Handle the user attempting to merge 2 accounts together
+        elif data.form == "mergeaccount":
+            if isinstance(data.passwd, unicode):
+                data.passwd = data.passwd.encode('utf-8')
+                
+            if isinstance(data.passwd2, unicode):
+                data.passwd2 = data.passwd2.encode('utf-8')
+                
+            #Make the user enter their username so they know what they're doing.
+            if not user.username == data.username.lower():
+                return 'json:', dict(success=False, message=_("First username incorrect."), data=data)
+                
+            #Check whether the user's supplied password is correct
+            if not User.password.check(user.password, data.passwd):
+                return 'json:', dict(success=False, message=_("First password incorrect."), data=data)
+                
+            #Make the user enter their username so they know what they're doing.
+            if not User.objects(username=data.username2.lower()):
+                return 'json:', dict(success=False, message=_("Unable to find user by second username."), data=data)
+                
+            other = User.objects(username=data.username2.lower()).first()
+                
+            #Check whether the user's supplied password is correct
+            if not User.password.check(other.password, data.passwd2):
+                return 'json:', dict(success=False, message=_("Second password incorrect."), data=data)
+                
+            #Make them type "merge" exactly
+            if not data.confirm == "merge":
+                return 'json:', dict(success=False, message=_("Merge was either misspelled or not lowercase."), data=data)
+                
+            log.info("User %s merged account %s into %s.", user.username, other.username, user.username)
+            user.merge(other)
+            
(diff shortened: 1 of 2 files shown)


brave/core/account/controller.py
```diff
@@ -335,6 +335,42 @@ def post(self, **post):
...
+        #Handle the user attempting to merge 2 accounts together
+        elif data.form == "mergeaccount":
+            if isinstance(data.passwd, unicode):
+                data.passwd = data.passwd.encode('utf-8')
+                
+            if isinstance(data.passwd2, unicode):
+                data.passwd2 = data.passwd2.encode('utf-8')
+                
+            #Make the user enter their username so they know what they're doing.
+            if not user.username == data.username.lower():
```
jboning: If you want to deploy this before migrating the DB, might be worth doing a check with the original case and falling back to lower(). (Similar below.)

I don't know what your deployment/migration schedule is like, so this may be irrelevant.

brave/core/account/template/settings.html
```diff
@@ -348,6 +348,140 @@
...
+                        if ($('#username2').val().length <= 0) {
+                            $('#username2').addClass('error');
+                            $('#modal form input[name="username2"]').after(
+                                '<div class="alert alert-error">' +
+                                    '<strong>Username required</strong>' +
+                                '</div>'
+                            );
+                            error = true
+                        }
+                        if ($('#username').val() == $('#username2').val()) {
```
jboning, line 444: This check isn't duplicated on the python side (and it probably should be).

More generally, I'm sad to see all of these checks repeated in JS and Python. It would be great if we had a proper system for returning and displaying errors on particular fields. That's probably not in scope for this change, though.

brave/core/account/controller.py
```diff
@@ -335,6 +335,42 @@ def post(self, **post):
...
+                return 'json:', dict(success=False, message=_("Unable to find user by second username."), data=data)
+                
+            other = User.objects(username=data.username2.lower()).first()
+                
+            #Check whether the user's supplied password is correct
+            if not User.password.check(other.password, data.passwd2):
+                return 'json:', dict(success=False, message=_("Second password incorrect."), data=data)
+                
+            #Make them type "merge" exactly
+            if not data.confirm == "merge":
```
jboning: nit: !=

brave/core/account/controller.py
Acidity: Unless the server admin takes this specific change and merges it with a month+ old version of the rest of Core, the vast majority of users won't be able to log in anyways, and those that can won't be affected by this. I'm not sure it's necessarily worth adding complexity to the code for a situation that should never happen.

jboning: I fixed the able-to-log-in thing in 3567ada74ad661789f848b1877caf7fe14459f44. It would be nice if everything keeps working until the database is migrated. But, SOUND is migrated now, so I have no skin in the game; it's up to you guys.

brave/core/account/template/settings.html
Acidity, line 444: Fixed. Thanks.

brave/core/account/controller.py
```diff
@@ -335,6 +335,48 @@ def post(self, **post):
...
+        #Handle the user attempting to merge 2 accounts together
+        elif data.form == "mergeaccount":
+            if isinstance(data.passwd, unicode):
+                data.passwd = data.passwd.encode('utf-8')
+                
+            if isinstance(data.passwd2, unicode):
+                data.passwd2 = data.passwd2.encode('utf-8')
+                
+            #Make the user enter their username so they know what they're doing.
+            if not user.username == data.username.lower() and not user.username == data.username:
```
jboning: nit again: !=

jboning:
My only complaint is that I was confused about which username was going to last and which was going to be deleted. Looking at the form again, I could see that it said "merge into this account", which implies that the other account is going away.... but it will probably confuse some people.

Other than that it seems to work nicely!

jboning:
Actually, a weirdness: I had two keys for the same characters on my two different core accounts. After I merged, the keys page showed both keys, but only key showed the characters.

jboning:
Given that the "weirdness" I saw was a result of failing to load characters on a second account, I think this bit is good to go!

Issue #126, opened 2014-04-24
AJAX form behaviour consistency


jboning:
On the account creation form, hitting "enter" does not submit the form (but does display the password manager's dialog?). Hitting the button in the UI submits the form ajaxily, but does not trigger the password manager.

On the password recovery form, hitting "enter" submits the form as a normal (non-ajax) request, and then the user just sees the json result: http://i.imgur.com/EEnArW1.png

We need to audit for consistency of behaviour across all forms and across different browsers. Hopefully we can have appropriately abstracted form code so that this is hard to screw up in the future.

PR #141, opened 2014-04-30
Let enter submit more forms

jboning:
Password management is still a little busted though (in Firefox at
least--clicking submit does not trigger 1password, but hitting enter
does).

Addresses part of the cause for #126. A full audit and browser check would still be worthwhile.


jboning:
This doesn't cover things in the settings dialogs. Really the settings code design should be reworked.

Issue #139, opened 2014-04-30
Enforce password strength requirements on reset passwords.


Acidity:
Passwords reset using the new email system do not have strength requirements like when users sign up initially.

Acidity:
Should be fixed by PR #142, still needs to be tested.

jboning:
Do you need to test this still, or is this OK to be closed?

Acidity:
Well I've never personally tested it, but looking at the code I imagine it should be fine. I just don't have a mail server on my test machine to test it with.

jboning:
Gonna go ahead and close this then.

Issue #138, opened 2014-04-30
Enforce Password requirements on the "Change Password" dialog


Acidity:
The in account/settings.

I'll try to work on this by the way.

Acidity:
Fixed by PR #142.

PR #142, opened 2014-04-30
More password strength enforcement

jboning:
Adds password strength checking to the account recovery and normal
password change flows. Really this should be refactored to not have this
check repeated three times.


jboning:
fixes #138, fixes #139

Issue #31, opened 2013-10-21
Verify reasonable behaviour if a character is added to two accounts.


amcgregor:
Character "Foo" added to Bob's account and the same character added to Eve's account should be treated as two separate characters for the purposes of authentication and authorization, to prevent shenanigans such as character masquerading and access to protected resources for an attacking user.

amcgregor:
See #33 and #34 for specifics.

Issue #46, opened 2014-02-01
Issue with uniqueness.


amcgregor:
NotUniqueError: Tried to save duplicate unique keys (E11000 duplicate key error index: brave_core.e_v_e_entity.$i_1 dup key: { : 94284552 })

Module brave.core.key.model:86 in importChars

Issue #34, opened 2013-11-20
Gracefully handle adding of duplicate EVE API keys.


amcgregor:
Currently allowed, but it shouldn't be.  Model needs the key ID field unique, front-end needs to handle the validation error.  (As per non-numeric key IDs.)

Issue #33, opened 2013-11-20
Gracefully handle duplicate character imports.


amcgregor:
```
URL: http://auth.bravecollective.net:8088/key/
File '/home/amcgregor/apps/adam/lib/python2.7/site-packages/WebError-0.10.3-py2.7.egg/weberror/evalexception.py', line 431 in respond
  app_iter = self.application(environ, detect_start_response)
File '/home/amcgregor/apps/adam/lib/python2.7/site-packages/Beaker-1.6.4-py2.7.egg/beaker/middleware.py', line 73 in __call__
  return self.app(environ, start_response)
File '/home/amcgregor/apps/adam/lib/python2.7/site-packages/Paste-1.7.5.1-py2.7.egg/paste/registry.py', line 379 in __call__
  app_iter = self.application(environ, start_response)
File '/home/amcgregor/apps/adam/lib/python2.7/site-packages/Paste-1.7.5.1-py2.7.egg/paste/config.py', line 105 in register_config
  app_iter = application(environ, start_response)
File '/home/amcgregor/apps/adam/lib/python2.7/site-packages/Beaker-1.6.4-py2.7.egg/beaker/middleware.py', line 155 in __call__
  return self.wrap_app(environ, session_start_response)
File '/home/amcgregor/apps/adam/lib/python2.7/site-packages/WebCore-1.1.2-py2.7.egg/web/core/locale.py', line 242 in __call__
  return self.application(environ, start_response)
File '/home/amcgregor/apps/adam/lib/python2.7/site-packages/WebCore-1.1.2-py2.7.egg/web/auth/middleware.py', line 103 in __call__
  result = self.application(environ, our_start_response)
File '/home/amcgregor/apps/adam/lib/python2.7/site-packages/WebCore-1.1.2-py2.7.egg/web/core/templating.py', line 84 in __call__
  result = self.application(environ, start_response)
File '/home/amcgregor/apps/adam/lib/python2.7/site-packages/WebCore-1.1.2-py2.7.egg/web/core/application.py', line 151 in __call__
  content = self.root(web.core.request._current_obj())
File '/home/amcgregor/apps/adam/lib/python2.7/site-packages/WebCore-1.1.2-py2.7.egg/web/core/dialects/core.py', line 92 in __call__
  return last.__after__(part(*remaining, **data))
[...]

PR #144, opened 2014-04-30
Feature/fix duplicate characters

Acidity:
Resolves Issues #31, #33, #34, and #46.

### DIFF
--- a/brave/core/key/controller.py
+++ b/brave/core/key/controller.py
@@ -11,7 +11,7 @@ from marrow.util.convert import boolean
 from marrow.util.bunch import Bunch
 from mongoengine import ValidationError
 
-from brave.core.key.model import EVECredential
+from brave.core.key.model import EVECredential, EVECharacterKeyMask
 from brave.core.util.predicate import authorize, authenticated, is_administrator
 
 
@@ -33,7 +33,7 @@ class KeyIndex(HTTPMethod):
         
         #Delete any character that the key owner has registered, but no longer has a key for.
         for c in owner.characters:
-            if not c.credential_for(0):
+            if not c.credential_for(EVECharacterKeyMask.NULL):
                 c.delete()
 
         if request.is_xhr:
@@ -102,7 +102,6 @@ class KeyList(HTTPMethod):
         
         try:
             record.save()
-            record.pull()
             characters = []
             for character in record.characters:
                 characters.append(dict(identifier = character.identifier, name = character.name))
--- a/brave/core/key/model.py
+++ b/brave/core/key/model.py
@@ -31,7 +31,7 @@ class EVECredential(Document):
     key = IntField(db_field='k')
     code = StringField(db_field='c')
     kind = StringField(db_field='t')
-    mask = IntField(db_field='a', default=0)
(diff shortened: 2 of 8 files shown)


Acidity:
I've tested this personally, but would appreciate someone else testing to confirm functionality.

brave/core/key/controller.py
```diff
@@ -89,6 +89,10 @@ def post(self, **kw):
                 
         except ValueError:
             return 'json:', dict(success=False, message=_("Key ID must be a number."), field='key')
+            
+        #Prevent the user from adding a key that has already been added.
+        if EVECredential.objects(key=data.key):
```
jboning: This is racy--if two accounts add the same key at the same time, they can both pass this check and then both add objects. Better to make key unique and then check for failure at save time.

brave/core/key/template/list.html
```diff
@@ -89,8 +89,8 @@
                                         '</button>' +
                                     '</td>' +
                                 '</tr>'
-			).appendTo('table.table tbody');
-			console.log("ADDED TO TBODY");
+                        ).appendTo('table.table tbody');
+                        console.log("ADDED TO TBODY");
```
jboning: probably OK to clean this up

brave/core/key/template/list.html
```diff
@@ -140,6 +140,11 @@
     <tr data-id="${record.id}">
         <td class="key-id"><i class="fa fa-fw fa-${'check' if record.verified else 'times'}"></i> ${record.key}</td>
         <td class="key-code"><span class="ellipsis">${record.code}</span></td>
+        % if record.violation == "Character":
+        <td class="security-violation"><font color="red">Character Security Violation.</font></td>
```
jboning: I'd prefer a more descriptive error message

brave/core/key/template/list.html
```diff
@@ -140,6 +140,11 @@
     <tr data-id="${record.id}">
```
jboning, line 140: is row() actually used? very similar stuff seems to be open-coded below.

jboning:
The row that gets added to the table ajaxily does not include the violation cell

jboning:
There's a complete lack of feedback when submitting an API key that's already on a different account. #145 :/

brave/core/key/model.py
```diff
@@ -177,12 +185,18 @@ def pull(self):
...
         
+        allCharsOK = True
+
         for char in result.characters.row:
             if 'corporationName' not in char:
                 log.error("corporationName missing for key %d", self.key)
                 continue
             
-            implementation(char)
+            if implementation(char) == (None, None, None):
```
jboning: This isn't the only place that calls pull_minimal -- grep for other call sites and fix those too.

jboning: On second though, hold off on fixing these. I'm about to tear up and refactor these functions in a big way, so don't worry about it.

Acidity:
Accidentally pushed this too early. Still has issues.

Acidity:
Still no feedback when attempting to add a key that's taken, but I'd rather leave that to another PR.

Acidity:
Replacing with new pull request.

PR #155, opened 2014-05-01
Makes EVE API key access and usage more readable.

Acidity:
No longer requires the use of bitwise operators in areas other than the key subdirectory, adds functions to check whether a certain mask supports any of the various eve api functions, and more!

### DIFF
--- a/brave/core/character/model.py
+++ b/brave/core/character/model.py
@@ -204,7 +204,7 @@ class EVECharacter(EVEEntity):
         lowest = None
         lowest_count = None
         for candidate in candidates:
-            bc = bin(candidate.mask).count('1')
+            bc = candidate.mask.number_of_functions()
             if lowest_count is None or bc < lowest_count:
                 lowest, lowest_count = candidate, bc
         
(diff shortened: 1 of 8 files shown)


Acidity:
Still want to double check that I hit all of the places that this affects before merging.

jboning:
https://github.com/bravecollective/core/blob/develop/brave/core/character/model.py#L199
https://github.com/bravecollective/core/blob/develop/brave/core/util/signal.py#L54

brave/core/character/model.py
```diff
@@ -199,7 +199,7 @@ def titlesort(i):
     def credential_for(self, mask):
         """Return the least-permissive API key that can satisfy the given mask."""
         
-        candidates = [i for i in self.credentials if not mask or not i.mask or i.mask & mask == mask]
+        candidates = [i for i in self.credentials if not mask or not i.mask or i.mask.has_access(mask)]
```
jboning, line 202: Below, `bin(candidate.mask).count('1')` is still going to break

jboning:
https://github.com/bravecollective/core/blob/develop/brave/core/group/acl.py#L191 one more I think. sorry for not catching it the first time around :(

brave/core/key/controller.py
```diff
@@ -102,7 +102,6 @@ def post(self, **kw):
         
         try:
             record.save()
-            record.pull()
```
jboning, line 105: looking again, I think maybe we do want to have this here. The pull triggered by the save will be asynchronous, so we won't necessarily have the character list ready to return to the web interface.

PR #157, opened 2014-05-02
Allows core admins to modify anyone's apps.

Acidity:
Allows core administrators to control which applications are allowed to show up in their application list, as well as modifying attributes of applications whose original owner may have left.


jboning:
It's super confusing that these classes are called OwnApplicationList and OwnApplicationInterface. I can't tell if the plan was to have a separate admin interface for applications?

As an admin it would be useful to see more things on the main table, like who created the app and the app's url.

Acidity:
Known (minor) bug: Admins that delete an application that is theirs will only have the row they clicked delete on disappear AJAXily. Refreshing resolves.

PR #159, opened 2014-05-02
Adds a check to ensure that the key being used to pull the character data...

Acidity:
...has a sufficient key mask.


jboning:
`info` starts with data from an APIKeyInfo call, so even a 0-permission key will pull basic things like character name, alliance, corporation (this is what the old pull_minimal() function did).

Acidity:
Deprecated by PR #163.

PR #160, opened 2014-05-02
Resolves some issues involved with deleting characters.

Acidity:


brave/core/character/model.py
```diff
@@ -209,3 +209,15 @@ def credential_for(self, mask):
                 lowest, lowest_count = candidate, bc
         
         return lowest
+        
+    def delete(self):
+        #If this character is the primary character for the account, wipe that field for the user
+        if self == self.owner.primary:
+            self.owner.primary = None
```
jboning, line 216: you'll need to actually save the change :)

PR #164, opened 2014-05-04
Feature/fix duplicate characters

Acidity:
Replaces #144.

### DIFF
--- a/brave/core/account/model.py
+++ b/brave/core/account/model.py
@@ -97,7 +97,6 @@ class User(Document):
         """Consumes other and takes its children."""
         
         other.credentials.update(set__owner=self)
-        other.credentials.update(set__violation="False")
         other.characters.update(set__owner=self)
         
         LoginHistory.objects(user=other).update(set__user=self)
--- a/brave/core/account/model.py
+++ b/brave/core/account/model.py
@@ -97,6 +97,7 @@ class User(Document):
         """Consumes other and takes its children."""
         
         other.credentials.update(set__owner=self)
+        other.credentials.update(set__violation="False")
         other.characters.update(set__owner=self)
         
         LoginHistory.objects(user=other).update(set__user=self)
--- a/brave/core/key/controller.py
+++ b/brave/core/key/controller.py
@@ -103,12 +103,12 @@ class KeyList(HTTPMethod):
         
         try:
             record.save()
-            characters = []
-            for character in record.characters:
-                characters.append(dict(identifier = character.identifier, name = character.name))
-
             #Necessary to guarantee that the pull finished before returning.
             record.pull()
+            characters = []
+            for character in record.characters:
+                characters.append(dict(identifier = character.identifier, name = character.name)
+            
 
             if request.is_xhr:
                 return 'json:', dict(
--- a/brave/core/key/controller.py
+++ b/brave/core/key/controller.py
@@ -107,7 +107,7 @@ class KeyList(HTTPMethod):
             record.pull()
             characters = []
             for character in record.characters:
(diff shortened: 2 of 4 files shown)


brave/core/account/model.py
```diff
@@ -97,6 +97,7 @@ def merge(self, other):
         """Consumes other and takes its children."""
         
         other.credentials.update(set__owner=self)
+        other.credentials.update(set__violation="False")
```
jboning: Here's a crazy scenario: user1 has a single-character key for Character One; user2 has a single-character key for Character Two; user3 has a full-account key covering both Character One and Character Two (and since this key has duplicate character, this key gets violation=True). Merging user2 and user3 shouldn't automatically unset the violation on the key, since there's still a conflict with user1.

We may also potentially want to clear violations on the first user's credentials (if the keys were added in the opposite order).

brave/core/key/controller.py
```diff
@@ -122,6 +128,11 @@ def post(self, **kw):
                         success = False,
                         message = _("Validation error: one or more fields are incorrect or missing."),
                     )
+        except NotUniqueError:
```
jboning, line 130: This is redundant with the check on line 101

brave/core/key/model.py
```diff
@@ -35,6 +36,7 @@ class EVECredential(Document):
     verified = BooleanField(db_field='v', default=False)
     expires = DateTimeField(db_field='e')
     owner = ReferenceField('User', db_field='o', reverse_delete_rule='CASCADE')
+    violation = StringField(db_field='s')
```
jboning, line 44: A comment here would be helpful. violation isn't as obvious as most of the other fields, an explanation of how it's used and possible values would be helpful for future maintenance.

brave/core/key/model.py
```diff
@@ -158,13 +167,19 @@ def pull(self):
...
             if 'corporationName' not in char:
                 log.error("corporationName missing for key %d", self.key)
                 continue
             
-            self.pull_character(char)
+            if not self.pull_character(char):
+                allCharsOK = False
         
+        if allCharsOK and self.violation == "Character":
+            self.violation = "False"
```
jboning: Weird. Can we use None to mean this instead? (or "")

brave/core/key/template/list.html
```diff
@@ -62,11 +62,20 @@
                         }
                         $.post('/key/', $('#modal form').serialize(), function(data)
                         {
+                            violationText = "";
+                            if(data.violation == "Character") {
+                                violationText = "<font color=\"red\">Character Security Violation.</font>";
+                            } else {
+                                violationText = "<font color=\"green\">The BIA is always watching...</font>";
+                            }
                             $(
```
jboning, line 71: I don't like this code very much. Probably should be cleaned up separately though.

Acidity:
Also, I don't know how making a field unique works for existing collections, it may cause problems if the database has 2 entries with the same keyid when this PR is deployed.

jboning:
ohgod, yeah, I don't know what happens when you add an index to the schema later. ???? research time I guess. (But you're right that the first check makes that safe, so maybe not to worry.)

Acidity:
I'd say that just to be safe, we should recommend a database check to verify that there are no entries in the Credentials collection that are duplicates.

brave/core/key/controller.py
```diff
@@ -106,6 +107,9 @@ def post(self, **kw):
             for character in record.characters:
                 characters.append(dict(identifier = character.identifier, name = character.name))
 
+            #Necessary to guarantee that the pull finished before returning.
+            record.pull()
```
jboning: This should be moved up to right below record.save(), so the character list is accurate

PR #165, opened 2014-05-04
Fix key mask

Acidity:
Fixes a minor bug with the return of key masks.


Acidity:
Whoops, this branch is above develop. Closing.

PR #169, opened 2014-05-04
Update the core.api call to update the character info from the EVE API.

Acidity:


Acidity:
Relies on PR #168 for proper functioning.

PR #170, opened 2014-05-04
Create an admin page for viewing all characters and their owners.

Acidity:
Potential problems: It's not paginated so it may take a while to load with thousands of characters.
No actions available on the admin page at the moment.

### DIFF
--- a/brave/core/character/template/list.html
+++ b/brave/core/character/template/list.html
@@ -115,8 +115,11 @@
                         <th class="span2 sortable name">Name</th>
                         <th class="span3 sortable corporation"><span class="line"></span>Corporation</th>
                         <th class="span3 sortable alliance"><span class="line"></span>Alliance</th>
+                        % if admin:
                         <th class="span3 sortable username"><span class="line"></span>User</th>
+                        % else:
                         <th class="span1 sortable actions"><span class="line"></span></th>
+                        % endif
                     </tr>
                 </thead>
                 <tbody>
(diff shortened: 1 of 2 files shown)


brave/core/character/template/list.html
```diff
@@ -93,10 +93,10 @@
     % if web.user.admin:
     <div id="tabs-wrapper">
         <div class="row-fluid">
-            <a class="span3 tab${' active' if not admin else ''}" href="/key/">
+            <a class="span3 tab${' active' if not admin else ''}" href="/character/">
```
jboning, line 96: haha what.

Acidity, line 96: Yeah, it originally sent the user over to the key management page...

brave/core/character/template/list.html
```diff
@@ -112,10 +112,14 @@
             <table class="table table-hover">
                 <thead>
                     <tr>
-                        <th class="span3 sortable name">Name</th>
-                        <th class="span4 sortable corporation"><span class="line"></span>Corporation</th>
-                        <th class="span4 sortable alliance"><span class="line"></span>Alliance</th>
+                        <th class="span2 sortable name">Name</th>
```
jboning: what's up with the changing span classes here?

Acidity: Ahhh, got this mixed up with bootstrap where the max width is 12 units. I'll get that fixed.

PR #171, opened 2014-05-04
Add the ability to mark applications as still in development.

Acidity:


brave/core/application/controller/browse.py
```diff
@@ -48,13 +48,13 @@ def index(self):
 class BrowseController(Controller):
     @authorize(authenticated)
     def index(self):
-        records = Application.objects()  # TODO: This needs to be ACL-filtered!
-        
-        # TODO: "Development"-only applications.
+        records = Application.objects(development=False or None)  # TODO: This needs to be ACL-filtered!
```
jboning: "development=False or None" ?

brave/core/application/form.py
```diff
@@ -20,6 +19,7 @@ def manage_form(action='/application/manage/'):
                     TextArea('description', L_("Description"), rows="6", class_='input-block-level'),
                     URLField('site', L_("Primary Site"), placeholder="http://", required=True, class_='input-block-level validate'),
                     EmailField('contact', L_("Primary Contact"), placeholder="user@example.com", class_='input-block-level'),
+                    TextField('development', L_("Development"), placeholder="True or False", class_='input-block-level'),
```
jboning, line 22: Janky. Can we use a checkbox or something instead?

brave/core/application/controller/browse.py
```diff
@@ -48,13 +48,13 @@ def index(self):
...
+        devRecords = Application.objects(development=True)
         
         return 'brave.core.application.template.list_apps', dict(
                 area = 'apps',
-                records = records
+                records = records,
+                devRecords = devRecords
```
jboning, line 57: nitpick: we usually use underscores here, don't we? (If I'm wrong and this is consistent, then that's fine.)

brave/core/application/form.py
Acidity, line 22: Read my comments!

jboning, line 22: Indeed, whoops. Okay I guess :/

brave/core/application/controller/browse.py
Acidity, line 57: hmm, I don't think so, and I was just following the convention of the code immediately surrounding it. the _ are for json dicts with messages I believe.

brave/core/application/controller/browse.py
Acidity: For compatibility with existing databases, if it was just False then those which had been added before the change would be non-visible to anyone on the available applications page. All new apps should have False by default though.

jboning: I see. I don't think this is going to accomplish what you want, though; I think you want `development__in=[False, None]`

Acidity: It worked perfectly in testing...

Acidity: Just kidding, you're right,

PR #172, opened 2014-05-05
Refresh script (fixes #88)

### LINKED ISSUE #88: EVE doesn't make key masks immutable; recheck masks periodically
API key masks and types need to be periodically checked in case someone submits a full API key then removes permissions from it.

jboning:


Acidity:
When trying to run the refresh script manually, I'm getting an ImportError, is there perhaps some additional setup required for this?

(brave)tyler@debianVM:~/brave/core$ python brave/core/scripts/refresh.py 
Traceback (most recent call last):
  File "brave/core/scripts/refresh.py", line 4, in <module>
    from brave.core import core_loadapp
ImportError: No module named core

jboning:
Do you have a virtualenv set up for core? Make sure you're running the right python. Though that wouldn't seem to lead to this. Also, have you done "python setup.py develop" or "pip install -e ." (I think that is the invocation)? That _might_ make a difference.

Acidity:
In a virtualenv (hence the (brave) next to the prompt, and I just ran both commands you suggested, and still getting the error, same with your case migration script. Might be just my set up though.

jboning:
Oh, derp. I've gotten used to running ../env/python manually. Weird. I think I saw a problem like this once, but I don't remember anything I did to make it go away.

Acidity:
I get the same problem with trying to run case_migration.py as well. I'll try this again tomorrow, though ideally we'd be able to get someone else to test and see if they have the same problem or not.

Acidity:
Yeah... problem was I wasn't running it in paster.

brave/core/scripts/refresh.py
```diff
@@ -0,0 +1,33 @@
...
+from brave.core import core_loadapp
+if __name__ == "__main__":
+    core_loadapp("config:"+sys.argv[1] if len(sys.argv) > 1 else None)
+
+import time
+
+from requests.exceptions import HTTPError
+
+from brave.core.key.model import EVECredential
+from brave.core.util.signal import validate_key
```
Acidity: Validate_key got deleted a while back, might want to fix this (replaced by pull iirc).

jboning: Whoops, sure enough. I'm really not on the ball with this one.

jboning:
What do you mean by running in paster? You shouldn't need to invoke paster yourself at all, core_loadapp is supposed to pull in all of the paster framework stuff.

Acidity:
Well, trying "python brave/core/scripts/refresh.py" leads to an import error, but doing "paster shell local.ini" then "import brave.core.scripts.refresh" works perfectly.

PR #174, opened 2014-05-05
Fix import error

jboning:
Really this is a questionable dependency; util ought not to rely on
model. This should be remedied. Moving the mask class to this
file would be perfectly reasonable.


Acidity:
Agreed, the mask class was originally only intended for EVECredentials, but it was usable for tokens and APICalls too, so I expanded it there. I'll submit a PR to resolve this.

PR #179, opened 2014-05-06
Add admin page for keys.

Acidity:

### DIFF
--- a/brave/core/key/controller.py
+++ b/brave/core/key/controller.py
@@ -78,11 +78,17 @@ class KeyList(HTTPMethod):
         
         if admin and not is_administrator:
             raise HTTPNotFound()
+            
+        credentials = user.credentials
+        if admin:
+            #Don't send the verification code for the API keys.
+            #See: The Great Brave API Key Leak of 2014
+            credentials = EVECredential.objects.only('violation', 'key', 'verified', 'owner')
 
         return 'brave.core.key.template.list', dict(
                 area = 'keys',
                 admin = admin,
-                records = user.credentials
+                records = credentials
             )
 
     @authorize(authenticated)
--- a/brave/core/key/controller.py
+++ b/brave/core/key/controller.py
@@ -82,7 +82,6 @@ class KeyList(HTTPMethod):
         credentials = user.credentials
         if admin:
             #Don't send the verification code for the API keys.
-            #See: The Great Brave API Key Leak of 2014
             credentials = EVECredential.objects.only('violation', 'key', 'verified', 'owner')
 
         return 'brave.core.key.template.list', dict(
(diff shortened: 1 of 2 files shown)


brave/core/key/template/list.html
```diff
@@ -148,18 +148,25 @@
...
-        <td class="key-code"><span class="ellipsis">${record.code | h}</span></td>
         % if record.violation == "Character":
         <td class="security-violation"><font color="red">Character Security Violation.</font></td>
         % else:
         <td class="security-violation"><font color="green">The BIA is always watching...</font></td>
         % endif
+        % if not admin:
+        <td class="key-code"><span class="ellipsis">${record.code | h}</span></td>
+        % else:
+        <td class="key-code" title="Thanks Draleth."><span class="ellipsis">Redacted.</span></td>
```
jboning: wat

brave/core/key/controller.py
```diff
@@ -78,11 +78,17 @@ def get(self, admin=False):
         
         if admin and not is_administrator:
             raise HTTPNotFound()
+            
+        credentials = user.credentials
+        if admin:
+            #Don't send the verification code for the API keys.
+            #See: The Great Brave API Key Leak of 2014
```
jboning: I think why is obvious enough, we don't need to go pointing fingers at specific people :P

brave/core/key/template/list.html
```diff
@@ -203,6 +210,9 @@
                         <th class="span3 sortable key-code"><span class="line"></span>${_("Verification Code")}</th>
                         <th class="span2 sortable characters hidden-tablet hidden-phone"><span class="line"></span>${_("Characters")}</th>
                         <th class="span3 sortable key-date hidden-phone"><span class="line"></span>${_("Added")}</th>
+                        % if admin:
+                        <th class="span3 sortable username"><span class="line"></span>${_("User")}</th>
```
jboning: the widths don't add up right in admin mode

Acidity: I'm not exactly sure what you mean by this (I don't use HTML often).

jboning: The bootstrap "span*" things are supposed to sum to something particular, right? With ?admin=1, the table is really wide (overflows the page for me).

Acidity: Well I thought it was supposed to be 12, but last time I modified a page to be 12 you questioned that :P

I'll go back and make both this and the other one have a width of 12, and see if that fixes it. I have no problems when viewing it in chromium as is though.

jboning: Oh. My question last time was "what's this about", and you had a good answer :P

PR #180, opened 2014-05-06
Lint

jboning:
This has two parts:
- travis configuration, which runs at a much more relaxed setting, to detect things that are almost certainly bugs.
- lint.py, which is a script to help detect style and other errors in your work. Run this before you send a PR :)

I pondered for a while and eventually gave up on defining a good ignore list. If lint.py warns you about something that you really think shouldn't be an error, feel free to add to the ignore line in setup.cfg.

### DIFF
--- a/brave/core/character/model.py
+++ b/brave/core/character/model.py
@@ -168,19 +168,6 @@ class EVECharacter(EVEEntity):
     credentials = ListField(ReferenceField(EVECredential, reverse_delete_rule=PULL), db_field='e', default=list)
     
     owner = ReferenceField('User', db_field='o', reverse_delete_rule=NULLIFY)
-
-    def update(self, alliance, corporation, race, bloodline, ancestry, gender, security, titles):
-        """A utility method to make sure that all important character properties are updated
-        together."""
-        self.alliance = alliance
-        self.corporation = corporation
-        self.race = race
-        self.bloodline = bloodline
-        self.ancestry = ancestry
-        self.gender = gender
-        self.security = security
-        self.titles = titles
-        self.roles = roles
     
     @property
     def tags(self):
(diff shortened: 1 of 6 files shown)


brave/core/character/model.py
```diff
@@ -168,19 +168,6 @@ class EVECharacter(EVEEntity):
     credentials = ListField(ReferenceField(EVECredential, reverse_delete_rule=PULL), db_field='e', default=list)
     
     owner = ReferenceField('User', db_field='o', reverse_delete_rule=NULLIFY)
-
-    def update(self, alliance, corporation, race, bloodline, ancestry, gender, security, titles):
```
jboning, line 172: I thought I cleaned this up from my key-update diff. Whoops.

Acidity:
If you run "python lint.py" without a command line argument, it throws an error. You might want to add a useful message telling the user to specify a branch. Other than that it looks good to me.

jboning:
Hm. If you have a tracking branch, it's supposed to use that; otherwise it's supposed to default to develop. I will take a look.

Acidity:
For what it's worth, I ran the script from FETCH_HEAD, so that may have something to do with it. I'll go grab the error for you though.

EDIT:
Traceback (most recent call last):
  File "lint.py", line 11, in <module>
    shell=True) or
  File "/usr/lib/python2.7/subprocess.py", line 544, in check_output
    raise CalledProcessError(retcode, cmd, output=output)
subprocess.CalledProcessError: Command 'git config branch.`git rev-parse --abbrev-ref HEAD`.merge' returned non-zero exit status 1

jboning:
Yeah, it was definitely busted for that case. (hurrah for testing my work >.>) But I think the new commit should fix it.

PR #184, opened 2014-05-07
Add script for deleting duplicate keys.

Acidity:

### DIFF
--- a/brave/core/scripts/fix_key_duplicates.py
+++ b/brave/core/scripts/fix_key_duplicates.py
@@ -1,0 +1,16 @@
+from brave.core.key.model import EVECredential
+
+def delete():
+    """ Deletes every duplicate key in the database."""
+    """ WARNING: DELETED EVERY INSTANCE OF THAT KEY."""
+    dups = []
+    
+    for credential in EVECredential.objects():
+        if len(EVECredential.objects(key=credential.key)) > 1:
+            for c in EVECredential.objects(key=credential.key):
+                dups.append(c)
+
+    print "Deleting {0} keys.".format(len(dups),)
+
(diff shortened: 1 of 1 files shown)


brave/core/scripts/fix_key_duplicates.py
```diff
@@ -0,0 +1,16 @@
...
+    """ Deletes every duplicate key in the database."""
+    """ WARNING: DELETED EVERY INSTANCE OF THAT KEY."""
+    dups = []
+    
+    for credential in EVECredential.objects():
+        if len(EVECredential.objects(key=credential.key)) > 1:
+            for c in EVECredential.objects(key=credential.key):
+                dups.append(c)
+
+    print "Deleting {0} keys.".format(len(dups),)
```
jboning: This count is going to be high by at least a factor of 2, since you're appending every duplicate key multiple times above. I think that on the whole the script still has the desired effect though :P

Acidity: Actually, the count is exactly as I intended, I wanted to output how many credentials were removed from the database. It's just the output is a bit confusing (since it's the number of keys in the database deleted rather than the number of unique keys deleted).

jboning: Right... and you're appending every credential twice. Let's give names... If `credential` A has a duplicate B, then we'll add both A and B to dups. And then later in the top-level loop, `credential` will be B, and we'll add both A and B to dups again.

Acidity: Ooops, you're right. Oh well, not really worth fixing, I only added that line so that the admin would know that it did something.

PR #185, opened 2014-05-07
Move the blacklist author to the config file.

Acidity:
Note: Admins will need to update their local.ini's for this to work.


jboning:
Adding the config field is probably a key for the todo-when-updating document.

Issue #5, opened 2013-10-04
Administrative account management.


amcgregor:
- [ ] Account list.
- [ ] Relevant links to browse filtered related data.
- [ ] Account CRUD views.
- [ ] Account CRUD actions.

PR #192, opened 2014-05-12
Add Admin management pages for users, characters, and keys.

Acidity:


Acidity:
This is still a work in progress. I just wanted to put this out here for comments and concerns.

Acidity:
Ready for review (finally)!

Acidity:
Resolves #5, #8.

jboning:
The KeyMask changes break API updating:

```
  File "/Users/josiah/eve/brave/core/brave/core/key/model.py", line 106, in pull_character
    elif self.mask.has_access(api.eve.CharacterInfo.mask):
  File "/Users/josiah/eve/brave/core/brave/core/util/eve.py", line 376, in has_access
    if self.mask & mask:
TypeError: unsupported operand type(s) for &: 'int' and 'NoneType'
```

Can we revert that change for this PR?

jboning:
I've decided to just go through this one chunk at a time, tear things off, and send them back to you as separate pull requests, sorry :/  I can see how each piece follows pretty naturally from the previous one, but putting them all together adds up to more than I'd like to review at once unless it's really necessary.

Can you talk a bit about the goal for the duplicate account detection? Are we just trying to be helpful and point out when a user is confused and created two accounts?

Acidity:
Deprecated in favor of all of those^.

PR #194, opened 2014-05-14
Fix bug where the popup information for security violations wasn't being...

Acidity:
... transmitted AJAXily.


jboning:
This row template isn't used?

PR #197, opened 2014-05-18
Add error messages for users.

Acidity:


Acidity:
Still looking for places which need this fixed for them.

jboning:
Cool. We may also want to think about ways to make this automatic, like using a jquery hook. It's odd because we're inconsistent currently... the login page has some nice ui built in, this uses an alert. Maybe it would be best to have an automatic alert, that individual post calls can opt out of if they want to do something prettier. Might not be easy, I haven't looked at the jquery request hook documentation yet... but something to think about.

PR #198, opened 2014-05-18
Fix a bug that broke the authorize buttons in the available tab,...

Acidity:
... and make the authorize button in the management tab work.

### DIFF
--- a/brave/core/application/template/list_apps.html
+++ b/brave/core/application/template/list_apps.html
@@ -60,7 +60,7 @@
                             <span class="subtext">${record.description | h}</span>
                         </td>
                         <td class="actions">
-                            <a href=${record.site | u}><button class="btn btn-small authorize">Authorize</button></a>
+                            <a href=${record.site}><button class="btn btn-small authorize">Authorize</button></a>
                         </td>
                     </tr>
                     % endfor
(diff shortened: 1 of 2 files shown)


jboning:
Hmmm, why is this (a) necessary and (b) okay? I'm really uncomfortable removing the escaping, and curious why it's not working and why this fixes it. I'd really like to find a better way to fix whatever the problem is.

Acidity:
This bug is actually live on SOUND's auth if you want to check it there, but with the escaping the buttons direct to (in SOUND's case) https://auth.of-sound-mind.com/application/browse/http%3A%2F%2Fknees.of-sound-mind.com%2F

jboning:
Hahah, I see. Random ideas:
- We should at the very least use the "h" filter.
- Ensuring http/https would be good, to avoid "javascript:"
- Maybe the safest thing would be another route that 302s the user to the right place

Acidity:
@jboning I'm actually going to leave this one to you, as you seem more invested in it.

jboning:
Okay, will try to hit this tonight or otherwise soon.

Acidity:
http://docs.makotemplates.org/en/latest/filtering.html < Lquid wanted me to send this on to you

PR #199, opened 2014-05-18
Fix issue with deleting characters when deleting a key.

Acidity:


jboning:
This seems a little convoluted. Can we just check `len(char.credentials) > 1`, or maybe `len([c for c in char.credentials if c.id != self.id]) > 0`? `credential_for_excluding` doesn't seem like a generally useful abstraction.

PR #204, opened 2014-05-23
Retain characters when their last API key is deleted.

Acidity:


jboning:
Cool, I like this. Just fix the bug (oh holy shit the tests caught a bug!!!) and you'll be good to go.

PR #205, opened 2014-05-23
Fix a bug that broke the authorize buttons in the available tab, and mak...

Acidity:
...e the authorize button in the management tab work.


Acidity:
Closing because I'm bad.

PR #207, opened 2014-05-23
fix misuse of "| u"

jboning:
(because I am a derp)

There are definitely further security improvements to be made here
(e.g. avoiding "javascript:" links). I'll send another change to fix
that shortly.


Acidity:
By the way, the site field for Applications uses Mongo's URLField, so checking for javascript: might not be necessary.

jboning:
Given the wackiness of URLField, I've added our own enforcement of http/https for site urls.

PR #211, opened 2014-05-23
The straight-template-relocation portion of #192

jboning:
Includes a bit of the actual code modification as well (addition of broken links) but I'm okay with that. (It's not like adding another one or two links that don't go anywhere will make a difference anyway....)


Acidity:
Replaced by #214.

PR #213, opened 2014-05-23
Fix a couple bugs in the delete pre leak script because I'm bad.

Acidity:
Thanks for noticing the bugs Pax!

### DIFF
--- a/brave/core/scripts/delete_pre_leak_keys.py
+++ b/brave/core/scripts/delete_pre_leak_keys.py
@@ -5,12 +5,10 @@ def delete(delete=False):
     
     x = 0
     
-    for credential in EVECredential.objects():
-        if len(EVECredential.objects(key=credential.key)) > 1:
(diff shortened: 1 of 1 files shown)


jboning:
the badness is mutual

brave/core/scripts/delete_pre_leak_keys.py
```diff
@@ -5,12 +5,10 @@ def delete(delete=False):
...
-            for c in EVECredential.objects(key__lt=3283828):
-                x += 1
-                print c.key
-                if delete:
-                    c.delete
+    for c in EVECredential.objects(key__lt=3283828):
+        x += 1
+        print c.key
+        if delete:
+            c.delete
```
jboning: while we're here, does `c.delete` actually work, or do you need to call it? (`c.delete()`)

Acidity: I don't know what you're talking about... :P

PR #215, opened 2014-05-23
Add duplicate account detection to core.

Acidity:
From #192.

### DIFF
--- a/brave/core/account/authentication.py
+++ b/brave/core/account/authentication.py
@@ -114,6 +114,18 @@ def authenticate(identifier, password):
     # Record the fact the user signed in.
     LoginHistory(user, True, request.remote_addr).save()
     
+    # Update the user's host
+    user.host = request.remote_addr
+    
+    # Check for other accounts with this IP address
+    if len(User.objects(host=request.remote_addr)) > 1:
+        # Quite possibly the worst code ever
+        for u in User.objects(host=request.remote_addr):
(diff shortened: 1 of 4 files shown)


brave/core/account/authentication.py
```diff
@@ -114,6 +114,18 @@ def authenticate(identifier, password):
     # Record the fact the user signed in.
     LoginHistory(user, True, request.remote_addr).save()
     
+    # Update the user's host
+    user.host = request.remote_addr
+    
+    # Check for other accounts with this IP address
+    if len(User.objects(host=request.remote_addr)) > 1:
+        # Quite possibly the worst code ever
+        for u in User.objects(host=request.remote_addr):
```
jboning, line 123: presumably all the other objects with this IP are already duplicates of each other... it's probably okay to just add_duplicate(user, other) for all others?

Acidity, line 123: Originally, I figured that was a bad assumption, but now that I think about it, since the host field doesn't get added in this update anyways we can probably make that assumption.

brave/core/account/authentication.py
```diff
@@ -114,6 +114,18 @@ def authenticate(identifier, password):
...
+            for z in User.objects(host=request.remote_addr):
+                User.add_duplicate(u, z, True)
```
jboning: style nit: it's usually clearer to pass a boolean as a keyword than a positional argument, so a reader can tell what it means

PR #216, opened 2014-05-23
Add account details page!

Acidity:
From #192.


jboning:
How much of this template is copied from the dashboard template? Would it be reasonable to share code?

Acidity:
The formatting style is borrowed from the dashboard template, but other things were thrown in to replace a lot of the content from the dashboard.
You could probably use an <%inherit file%> from a common base for these 2, but I'm not sure it's worth the effort.

PR #217, opened 2014-05-23
Remove the unnecessary functions in the EVEKeyMask classes, add a...

Acidity:
...function method to them, and update all references to them to now refer to APICall objects.

Part of #192.

### DIFF
--- a/brave/core/util/eve.py
+++ b/brave/core/util/eve.py
@@ -171,6 +171,8 @@ class APICall(Document):
     description = StringField()
     _mask = IntField(db_field="mask")
     group = IntField()
+    # Allow for the name of an APICall to be different from the URI in certain circumstances
+    uriname = StringField()
     
     @property
     def mask(self):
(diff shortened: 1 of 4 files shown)


Acidity:
Part of #192

jboning:
Same problem as before here, CharacterInfo has a mask of None

brave/core/util/eve.py
```diff
@@ -171,13 +171,14 @@ class APICall(Document):
     description = StringField()
     _mask = IntField(db_field="mask")
     group = IntField()
+    # Allow for the name of an APICall to be different from the URI in certain circumstances
+    uriname = StringField()
```
jboning, line 175: did this end up getting used at all?

brave/core/util/eve.py
```diff
@@ -350,11 +354,19 @@ def populate_calls(force=False):
         APIGroup(row.groupID, row.name, row.description).save()
     
     for row in result.calls.row:
-        APICall(row.type.lower()[:4] + '.' + row.name,
-            type_mapping[row.type],
-            row.description,
-            row.accessMask,
-            row.groupID).save()
+        if row.type.lower()[:4] != 'char' or row.name != 'CharacterInfo':
```
jboning: Not a priority, but if you are updating the code again, I think it's clearer to invert the structure of this. Detect and handle the special case first, then put the general case in the "else". (This is a reasonably common pattern; one reason is that it extends more nicely to several special cases.)

jboning:
The api.eve.CharacterInfo vs api.char.CharacterInfo discrepancy is still weird, but yeah, this seems as good a workaround as we'll get :| Maybe someday we'll think of something better.

PR #219, opened 2014-05-23
Add Key detail page, and key mask detail page.

Acidity:
Part of #192.

### DIFF
--- a/brave/core/key/template/keyDetails.html
+++ b/brave/core/key/template/keyDetails.html
@@ -1,0 +1,110 @@
+## encoding: utf-8
+
+<%inherit file="brave.core.template.master"/>
+
+<%block name="title">${_("EVE API Keys")}</%block>
+
+<%block name="header">
+    ${parent.header()}
+    
+    <style>
+        .modal-body form { margin-bottom: 0; }
+        .modal-body form input.input-block-level:last-child { margin-bottom: 0; }
+        table.table { table-layout: fixed; width: 100%; }
+        span.ellipsis { display: inline-block; width: 100%; overflow: hidden !important; white-space: nowrap !important; text-overflow: ellipsis !important; word-break: break-all !important; word-wrap: break-word !important; }
+    </style>
+</%block>
+
+<%block name="post">
+    ${parent.post()}
+    
+    <script type="text/javascript" charset="utf-8">
+        $(function()
+        {
+
+            $('time').timeago();
+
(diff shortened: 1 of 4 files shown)


brave/core/key/template/keyDetails.html
```diff
@@ -0,0 +1,110 @@
...
+        <b>Type:</b> ${record.kind}<br/>
+        % if record.kind == 'Character' or record.kind == 'c':
+        <b>Characters:</b> 
+        <% 
+        x = 1 
+        char_line = ""
+
+        for c in record.characters:
+            if c.alliance:
+                char_line += "<a href=\"/character/" + str(c.id) + "\">[" + c.alliance.short + "] " + c.name + "</a>"
```
jboning, line 82: I fear this whole section

Acidity, line 82: Hmmm, Why?

jboning, line 82: In particular, that the character name and alliance short-thing are unescaped (I don't know what CCP allows or denies here, but better to be safe than sorry).

On top of that it's a little odd to be building up html by hand in python inside the template file, but maybe there's not much of a better way to do it, I'm not completely sure here

Acidity, line 82:
```
a. EVE Online account usernames must contain at least four alphanumeric characters,
may not contain spaces or unconventional characters and must not be vulgar, 
obscene, ethnically, racially or sexually offensive or otherwise inappropriate.

a. Valid player character names:
Must be at least 4 characters.
Cannot exceed 24 characters.
May contain the characters A-Z, a-z, 0-9, and single quotation. (Corporation names may also include minus and dot characters.)
Space or single quotation characters are not allowed as the first or last character in a name.
Must be unique.
```

Doesn't look like there's anything abusable in there, but if you want me to add the escaping that is certainly possible.

brave/core/key/template/maskTable.html
```diff
@@ -0,0 +1,49 @@
...
+            </tr>
+        </thead>
+        <tbody>
+            % for func in all_functions:
+                % if x % 6 == 0:
+                    <tr>
+                % endif
+                <td>${colors[x]}${all_functions[x].name[5:]}
+                % if x % 6 == 5:
+                    </tr>
```
jboning, line 42: this won't necessarily close the last table row. It probably works fine, might be nice to clean up

PR #220, opened 2014-05-23
Add search functionality.

Acidity:
Part of #192.

### DIFF
--- a/brave/core/admin/controller.py
+++ b/brave/core/admin/controller.py
@@ -1,0 +1,124 @@
+# encoding: utf-8
+
+from __future__ import unicode_literals
+
+from datetime import datetime
+
+from web.auth import user
+from web.core import Controller, HTTPMethod, request
+from web.core.locale import _
+from web.core.http import HTTPFound, HTTPNotFound
+
+from brave.core.character.model import EVECharacter, EVECorporation, EVEAlliance
+from brave.core.group.model import Group
+from brave.core.group.acl import ACLList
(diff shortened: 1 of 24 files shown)


jboning:
I'd suggest breaking up the search input UI and endpoint to be separate forms and endpoints. Clearer for the user that you can only search one kind of thing at a time, and I think clearer reading the code, too.

brave/core/admin/controller.py
```diff
@@ -0,0 +1,124 @@
...
+
+class AdminController(Controller):
+    """Entry point for the Search RESTful interface."""
+
+    #index = AdminInterface()
+
+    def index(self):
+        return 'brave.core.admin.template.search', dict(area='admin')
+    
+    def search(self, Character=None, charMethod=None, Alliance=None, Corporation=None, group=None, KeyID=None,
```
jboning: curious why the uppercased characters (usually uppercase is used to indicate class names in python)

also, "group" here does not match "Group" in the template

jboning:
oh, also, the select inputs for how to match character/username are kind of confusing

jboning:
Did you mean to add new template files?

I did not mean to suggest breaking up the search forms onto separate pages, sorry :( just making the difference search types visibly different forms. Don't feel obligated to change it though, I'm happy to try some different designs myself after landing.

Acidity:
errr, yeah I forgot the templates...

Also, I think the separate pages works better from a design standpoint now that I see it when it's actually done, but that may just be me.

PR #223, opened 2014-05-24
Character and account page tweaks

jboning:
- widen the main section and shrink the sidebar
- make section headers consistent
- add detail to page titles
- demote prominence of primary character on account page

(as usual feel free to argue with any and all of these)


Acidity:
My only suggestion would be adding the primary character's corp titles to the account details page again, so that admins can more easily identify leadership/important people.

jboning:
Re-added. I'd wanted to emphasize the contrast between character and account pages, but you make a good point.

PR #231, opened 2014-05-25
Fix a bug that prevented the deletion of characters that had no owner.

Acidity:


jboning:
just curious, where is delete() being used these days?

Acidity:
In paster shell when something gets messed up and needs to get deleted manually.

PR #232, opened 2014-05-25
Change the Development field into an actual checkbox

paxswill:
Figured out how to get marrow's checkboxes working.


jboning:
rockin', thanks!

Acidity:
Does this save the value of the development field? There was a bug before this change where the dev field had to be filled in every time the app settings were changed.

jboning:
Yeah (or at least it worked when I tested it)

paxswill:
Mhm, fixed that bug.

Issue #131, opened 2014-04-26
API access to non-primary characters


jboning:

Acidity:
This already exists I believe... Only thing is that applications have no way to know which characters are owned by the same core account and as such have to treat each character as a separate user.

paxswill:
Couldn't this be resolved by [allowing access](https://github.com/bravecollective/core/blob/develop/brave/core/api/eve.py#L34) to the [account.Characters](http://wiki.eve-id.net/APIv2_Account_Characters_XML) API endpoint? The security implications of allowing access to that endpoint seem fairly minor (and an application still has to have that access mask). Really, I'm not having a whole lot of luck thinking of many application uses for the account.\* calls besides Characters.

mglinski:
No, then you would only see characters on the account whose API key was selected for use on the proxy. There needs to be a way to select multiple characters to allow access to on an authorized app and a dedicated core endpoint for getting all those characters' /info calls at once for an authorized app.

PR #233, opened 2014-05-26
Add support for multiple characters in application grants

paxswill:
I think this should address #131. Backwards compatibility with old consumers of the api.core.info endpoint is maintained by using (in order of preference) the primary character, or the first character alphabetically (well, really whatever the db returns as the first character, I left that undefined) for the info returned by the old API.

I tried to work on the styling of the authorization window, and I think it looks _alright_, but it could do with some more work.

~~An area I'm still looking to improve on is allowing users to edit existing grants. This means users don't have to revoke and then re-authorize an app just to add (or remove) characters.~~ Done as of paxswill@8a823ff2d8c2b398c8bc7b38e09519aac951f0c8.

### DIFF
--- a/brave/core/api/core.py
+++ b/brave/core/api/core.py
@@ -138,13 +138,7 @@ class CoreAPI(SignedController):
         token = ApplicationGrant.objects.get(id=token, application=request.service)
 
         # Step 2: Update info about the characters from the EVE API
-        # Support old grants that only have a single character.
-        # Could be removed after performing a database migration.
-        try:
-            characters = token.characters
-        except AttributeError:
-            characters = [token.character]
-
+        characters = token.characters
         characters_info = []
         for character in characters:
             mask, key = character.credential_multi_for((api.char.CharacterSheet.mask,
--- a/brave/core/api/core.py
+++ b/brave/core/api/core.py
@@ -136,18 +136,53 @@ class CoreAPI(SignedController):
         
         # Step 1: Get the appropraite grant.
         token = ApplicationGrant.objects.get(id=token, application=request.service)
-        character = token.character
-        
-        # Step 2: Update info about the character from the EVE API
-        mask, key = character.credential_multi_for((api.char.CharacterSheet.mask,
-                                                    api.char.CharacterInfoPublic.mask, EVECharacterKeyMask.NULL))
-        
-        # User has no keys registered.
-        if not key:
-            return None
-            
-        key.pull()
-        
+
+        # Step 2: Update info about the characters from the EVE API
+        # Support old grants that only have a single character.
+        # Could be removed after performing a database migration.
+        try:
+            characters = token.characters
+        except AttributeError:
+            characters = [token.character]
+
+        characters_info = []
+        for character in characters:
+            mask, key = character.credential_multi_for((api.char.CharacterSheet.mask,
+                                                        api.char.CharacterInfoPublic.mask, EVECharacterKeyMask.NULL))
+
+            # Character has no keys registered.
+            if not key:
+                continue
+            key.pull()
+
+            # Fill in the character info
+            character_info = {
+                'character': {
+                    'id': character.identifier,
+                    'name': character.name,
+                },
+                'corporation': {
+                    'id': character.corporation.identifier,
+                    'name': character.corporation.name,
+                },
+                'primary': token.user.primary == character,
+            }
+            if character.alliance:
+                character_info['alliance'] = {
+                    'id': character.alliance.identifier,
+                    'name': character.alliance.name,
+                }
+            characters_info.append(character_info)
+
+        # Backwards compatibility for apps using the old API
+        if len(characters) == 1:
+            character = characters[0]
+        elif token.user.primary in characters:
+            character = token.user.primary
+        else:
+            # LOL, pick one!
+            character = characters[0]
+
         # Step 3: Match ACLs.
         tags = []
         for group in Group.objects(id__in=request.service.groups):
@@ -158,6 +193,7 @@ class CoreAPI(SignedController):
                 character = dict(id=character.identifier, name=character.name),
                 corporation = dict(id=character.corporation.identifier, name=character.corporation.name),
                 alliance = dict(id=character.alliance.identifier, name=character.alliance.name) if character.alliance else None,
+                characters = characters_info,
                 tags = tags,
                 expires = None,
                 mask = token.mask
--- a/brave/core/api/core.py
+++ b/brave/core/api/core.py
@@ -6,7 +6,7 @@ from operator import __or__
 
 from web.core import request, response, url, config
 from web.auth import user
-from mongoengine import Q, DoesNotExist
+from mongoengine import Q
 from marrow.util.url import URL
 from marrow.util.object import load_object as load
 from marrow.util.convert import boolean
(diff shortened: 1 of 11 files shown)


brave/core/api/core.py
```diff
@@ -136,28 +136,58 @@ def info(self, token):
...
+                character_info['alliance'] = {
+                    'id': character.alliance.identifier,
+                    'name': character.alliance.name,
+                }
+            characters_info.append(character_info)
+
+            # Step 3: Match ACLs.
+            for group in Group.objects(id__in=request.service.groups):
+                if group.evaluate(token.user, character):
+                    tags.append(group.id)
```
Acidity: Important distinction: Characters, not users, have groups. You should keep characters' groups separate.

brave/core/api/core.py
```diff
@@ -136,28 +136,58 @@ def info(self, token):
...
-            if group.evaluate(token.user, character):
-                tags.append(group.id)
-        
+        for character in characters:
+            mask, key = character.credential_multi_for((api.char.CharacterSheet.mask,
+                                                        api.char.CharacterInfoPublic.mask, EVECharacterKeyMask.NULL))
+
+            # Character has no keys registered.
+            if not key:
+                continue
```
Acidity, line 145: How do you wish to handle when a character in the token gets removed from the user's account? You could 1) delete the token or 2) remove the character from the token. Just checking that there is a key is insufficient, as that key could be registered to a different account. (See brave/core/character/model.py function detach())

brave/core/scripts/multiple_characters_grant_migration.py
```diff
@@ -0,0 +1,16 @@
...
+    """Migrate from single-character application grants to multi-character ones
+    """
+
+    count = 0
+    for grant in ApplicationGrant.objects(character__ne=None):
+        print '{g.application.name} - {g.character.name}'.format(g=grant)
+        count += 1
+        if not dry_run:
+            grant.characters = [grant.character]
+            grant.character = None
```
Acidity, line 13: It seems like it might be a good idea to leave the character be, especially if you wish to retain backwards compatibility of core.info.

Acidity:
Is there a way for applications to specify how many characters they're willing to accept? Some applications (like mumble) don't have an easy way to handle multiple characters.

brave/core/scripts/multiple_characters_grant_migration.py
paxswill, line 13: Backwards compatibility is already maintained in core.info by supplying the same information in the same format as the old core.info calls. The only added information is a `characters` key, with an array of characters. I've backed out (as of paxswill@447286a8b324f602dbb60ee8e46848e74e8eae45) the `tags` key be a union of all of the characters' tags, and it is now limited to the same character as the rest of the old keys. Tag information for all characters (filtered by grant) is now available from the mappings in the `characters` array.

brave/core/api/core.py
paxswill, line 145: Hadn't thought of that, should be resolved in paxswill@8e6d155b6ad9be326a2d7aab89937d503fb4ad52

brave/core/application/controller/grant.py
```diff
@@ -28,6 +30,49 @@ def __init__(self, grant):
...
+    @authorize(authenticated)
+    def post(self, **kwargs):
+        if not request.is_xhr:
+            raise HTTPNotFound()
+
+        grant = self.grant
+        valid, invalid = edit_grant(grant).native(kwargs)
+
+        new_characters = []
+        print grant.user.characters
```
Acidity: Leftover testing code?

brave/core/application/controller/grant.py
```diff
@@ -28,6 +30,49 @@ def __init__(self, grant):
...
+        for key, value in valid.items():
+            try:
+                character = EVECharacter.objects.get(identifier=int(key))
+            except EVECharacter.DoesNotExist:
+                continue
+            if character in grant.user.characters and value:
+                print 'adding', character
```
Acidity: See above.

brave/core/application/controller/grant.py
```diff
@@ -28,6 +30,49 @@ def __init__(self, grant):
...
+                new_characters.append(character)
+            else:
+                print 'not adding', character, value
```
Acidity: See above.

brave/core/api/core.py
```diff
@@ -134,31 +134,81 @@ def info(self, token):
...
+            for group in Group.objects(id__in=request.service.groups):
+                if group.evaluate(token.user, char):
+                    char_tags.append(group.id)
+            character_info['tags'] = char_tags
+            if character_info['primary']:
+                character = char
+                tags = char_tags
+
+            characters_info.append(character_info)
+        else:
```
jboning, line 199: Is there a reason there's an else here and break on line 166, instead of just returning above? I found that structure kind of confusing.

brave/core/application/controller/grant.py
```diff
@@ -28,6 +30,44 @@ def __init__(self, grant):
...
+
+        grant = self.grant
+        valid, invalid = edit_grant(grant).native(kwargs)
+
+        new_characters = []
+        for key, value in valid.items():
+            try:
+                character = EVECharacter.objects.get(identifier=int(key))
+            except EVECharacter.DoesNotExist:
+                continue
```
jboning, line 59: would it be better to return failure instead?

brave/core/api/core.py
```diff
@@ -134,31 +134,81 @@ def info(self, token):
...
+            if character is None:
+                # There are multiple characters, and none of them are the primary character. Just picking one now.
+                character = token.characters[0]
+                tags = characters_info[0]['tags']
+
+            return dict(
+                    character = dict(id=character.identifier, name=character.name),
+                    corporation = dict(id=character.corporation.identifier, name=character.corporation.name),
+                    alliance = dict(id=character.alliance.identifier, name=character.alliance.name) if character.alliance else None,
+                    characters = characters_info,
```
jboning, line 210: What are your thoughts on adding this here vs creating a new endpoint? Both work, I'm just curious whether there's a reason to prefer this way.

jboning:
I played with the UI some. Take a look at 94d120d

brave/core/api/core.py
paxswill, line 210: No particular reason, I just did it this way to make adding multiple character support easier to add in to the SRP app (which was what I was testing with).

brave/core/application/controller/grant.py
paxswill, line 59: Possibly, but if it failed for some reason, they could re-add the character using the grant modification UI.

jboning, line 59: Yeah, definitely. I'm just worried that it won't be obvious when a character is dropped.

brave/core/api/core.py
jboning, line 210: my kneejerk reaction was to prefer a separate endpoint... but after stewing on it for a while, I haven't come up with a good reason why, and it does make sense to provide easy access to the primary character, so I'm happy to go with this.

jboning:
@paxswill, what do you want to do with this? land as-is and tweak UI separately, or hold for UI updates here? Also, I saw mention on jabber of another idea for managing multi-character grants... do you want to pursue that design some, or...?

paxswill:
Yeah, thinking about it more, better management of single-character grants might be the way to go. I think it would be simpler to implement, as I think it'd just be changing the authorization/login controller. Managing which apps have grants to which characters is easier, as the grant list should already handle them. The API wouldn't change, so no backwards compatibility concerns. It'd also simplify my usage in the SRP app, as now permissions have a straightforward mapping between users and pilots.

The downside is that it complicates things for end-user. I'm percolating an idea for how to make it simpler, but it's not all there yet.

Issue #229, opened 2014-05-25
Pull_character dies a horrible death if a character's ship name uses non-ASCII characters.


Acidity:
Ship Name: カリバチ™

Stack Trace:

```
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/home/core/src/brave/core/key/model.py", line 202, in pull
    if not self.pull_character(char):
  File "/home/core/src/brave/core/key/model.py", line 107, in pull_character
    info = api.eve.CharacterInfo(self, characterID=info.characterID)
  File "/home/core/src/brave/core/util/eve.py", line 244, in __call__
    data = xml(data)['eveapi']
  File "/home/core/local/lib/python2.7/site-packages/relaxml-0.1.3-py2.7.egg/relaxml/converter.py", line 34, in xml
    string = str(data)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 288-292: ordinal not in range(128)
```

jboning:
https://github.com/zachwill/relaxml/blob/master/relaxml/converter.py#L31 sigh.

jboning:
P.S. it would be great to have a test to prevent regressions :) I'll do this when I get a chance, unless you beat me to it

PR #237, opened 2014-05-26
Handle issue #229 better, as well as other errors which may occur when...

Acidity:
...making the call to the EVEAPI when pulling a character.

NOTE: This is not a solution to #229, just a workaround that mitigates the issues relating to it and other errors.

### DIFF
--- a/brave/core/key/model.py
+++ b/brave/core/key/model.py
@@ -92,8 +92,10 @@ class EVECredential(Document):
         from brave.core.character.model import EVEAlliance, EVECorporation, EVECharacter
         try:
             char = EVECharacter(identifier=info.characterID).save()
+            new = True
         except NotUniqueError:
             char = EVECharacter.objects(identifier=info.characterID)[0]
+            new = False
             
             if char.owner and self.owner != char.owner:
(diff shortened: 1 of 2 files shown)


brave/core/key/model.py
```diff
@@ -107,10 +109,17 @@ def pull_character(self, info):
...
-        if self.mask.has_access(api.char.CharacterSheet.mask):
-            info = api.char.CharacterSheet(self, characterID=info.characterID)
-        elif self.mask.has_access(api.char.CharacterInfoPublic.mask):
-            info = api.char.CharacterInfoPublic(self, characterID=info.characterID)
+        try:
+            if self.mask.has_access(api.char.CharacterSheet.mask):
+                info = api.char.CharacterSheet(self, characterID=info.characterID)
+            elif self.mask.has_access(api.char.CharacterInfoPublic.mask):
+                info = api.char.CharacterInfoPublic(self, characterID=info.characterID)
+        except:
```
jboning: Whoa! Almost never use bare "except:". This will catch a whole bunch of things you usually don't want to (like SystemExit). Most of the time you should catch specific exceptions, if you don't know what you want to catch then catch Exception.

PR #238, opened 2014-05-26
Prevent settings edits on other users' accounts

jboning:
It turns out this was actually okay because all of the edits require
password entry currently. But that was really nonobvious.

I made a pass through the other controllers checking auth stuff; there
are a lot of places where we 500 on not-logged-in users, but that's when
as a result of checking IDs. It would be nicer to present a better error
message, but I think we're okay for now.


Acidity:
Good catch

Issue #148, opened 2014-05-01
Add a general account details page.


Acidity:
Should include username, email address on file, etc. Should be at /account/index.html.

PR #246, opened 2014-05-31
Adding /account/ page where user can see his profile.

BlackVoid:


Acidity:
The account details page was never really intended to be user-accessible, as shown by tables shown being the admin-variant. Generally speaking, I would prefer that /account/ met the suggestions outlined in #148 where it's just a basic account specific page that doesn't really contain anything related to Eve (So things like username, email, complete login history, maybe even the buttons from the settings page (which seems superfluous to me...)). I'd like to get @jboning's opinion on this matter, as I'm obviously biased towards my own suggestion.

jboning:
This is a little redundant with the overview (home) page... maybe it would be better to add a summary of characters and keys to the overview? (The character column can be removed both there and on /account/<id>, but doing that cleanly might be tricky.)

That said, I don't really know what should go at /account/ vs /settings/. The settings interface should definitely be reworked to be friendlier. But I don't know what content should actually go here vs the settings page. Maybe it is better to leave the overview page as is, with the login history fairly prominent, and put a more complete summary here.

wtb UX designer

bargulg:
merged manually

PR #247, opened 2014-06-03
Csrf protection (fixes #92)

### LINKED ISSUE #92: Add xsrf protections

jboning:
This requires that all effectful requests be done via ajax, which I
think is already true. If we wish to support normal forms, that is
possible, but will require slightly more work.

The "light" base template is not quite as light as before. Oh well.

### DIFF
--- a/brave/core/controller.py
+++ b/brave/core/controller.py
@@ -2,11 +2,14 @@
 
 from __future__ import unicode_literals
 
+import os
+
+from binascii import hexlify
 from datetime import datetime, timedelta
 
-from web.core import Controller, HTTPMethod, config, session, request
+from web.core import Controller, HTTPMethod, config, session, request, response
 from web.auth import user
-from web.core.http import HTTPFound, HTTPNotFound
+from web.core.http import HTTPBadRequest, HTTPFound, HTTPNotFound
 from web.core.locale import set_lang, LanguageError, _
 from marrow.util.convert import boolean
 from marrow.util.url import URL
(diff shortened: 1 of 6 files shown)


Acidity:
I'm curious as to why you made the searches use GET instead of POST.

jboning:
The alternatives were doing the searches ajaxily, or making non-ajax post requests work with the csrf protection. This was less work >_>

brave/core/controller.py
```diff
@@ -125,6 +128,30 @@ class RootController(StartupMixIn, Controller):
...
+
+        return super(RootController, self).__call__(req)
+
+    def check_csrf(self):
+        # portions of the application explicitly opted out of CSRF protection.
+        if request.path_info_peek() == 'api':
+            return
+
+        if request.headers.get('X-CSRF'):
+            # the browser prevents sites from sending custom HTTP
```
Acidity, line 146: Which browsers does this apply to? All of them? Is this in the HTTP spec or do we need to limit the browsers we accept as valid for Core?

jboning, line 146: _handwave_ the paper's authors said so _handwave_

references for reading:
http://www.w3.org/TR/XMLHttpRequest/#the-setrequestheader%28%29-method
http://www.w3.org/TR/cors/#resource-preflight-requests
https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS

I haven't actually spent the time to sift through and grok these yet, this is as much a note to myself as anything.

Acidity:
I'd personally prefer if we extended this protection to normal Post's as well.

jboning:
Yeah. (_grumble grumble_.) I'm just dragging my feet.

For what it's worth, there's no reason for the searches _not_ to be GET. I know that keeping request parameters out of the url is all the style these days, but there's really no reason someone shouldn't be able to bookmark a search, or send a search to someone else.

Acidity:
Also, this is supposed to prevent links such as http://localhost:8080/admin/search/char?character=Acid&charMethod=contains&alliance=&corporation=&group=&submit=Search from working when clicked from another web site right?

jboning:
No, links like that are allowed, and should be, since GET requests should never cause side effects. (That's technically an oversimplification, since of course there is logging etc--the spec says "the GET and HEAD methods SHOULD NOT have the significance of taking an action other than retrieval", https://tools.ietf.org/html/rfc2616#section-9.1.1)

Acidity:
Ah ok, I thought you had enabled CSRF on GET and was just using that link to test your changes because forging a POST message requires more work.

PR #250, opened 2014-06-03
Add banning support to Core

Acidity:
Still a work in progress. Subject to change significantly.
TODO:
Test it.
(DONE) **_Verify that users that are banned can't log in**_
(DONE, NEEDS REASON ABILITY) Add banning button to the user detail pages (#192)
(DONE) Add banned search to the search page (192)
(DONE FOR USERS) Add banned detail to the user and character detail pages (192)
(DONE) Modify Schema for #204
(Deferred to later PR.) Add ability to unban.
(DONE) Check for enabled when verifying banned status.

Future PRs: Add ability to specify the reason for a ban, enable/disable subbans, maybe add a secret reason, add ability to unban a user, ban characters when added to Core on key with already banned character.


brave/core/account/controller.py
```diff
@@ -421,6 +426,21 @@ def get(self):
...
             account=self.user,
         )
+        
+    def post(self, ban=None, reason=None):
+        """ban = True bans the user, ban=False unbans the user, and ban=None leaves the user's ban status alone."""
+        
+        reason = 'test'
+        
+        if not reason:
+            return 'You must include a reason for this ban.'
```
jboning, line 436: does this work right?

jboning:
I'm curious for your reasoning behind the subban design. It seems more complicated than just banning characters and IPs individually. What's the win with this system?

brave/core/account/controller.py
Acidity, line 436: Yes, that is a feature to be added in a future PR.

Acidity:
As far as I can tell, there's about 3 different ways we could go about banning players. Before I discuss that, I want to point out an important detail for why I chose to implement banning this way: accounts are completely worthless from an access control perspective since anyone can create another account in ~5 seconds.
So the three main implementations are: 1) Make admins manually ban characters and IP addresses. This adds a significant amount of work for admins, as they must go through and ban all of the characters associated with the person they want banned. Additionally, it leads to the database being full of Bans that are logically the same ban, but technically separate. Additionally, this method prevents Core from being able to automatically ban new characters if it later finds out that that character is associated with a prior banned character, leading to a relatively ineffective banning solution. 
2) Have each character and IP address Ban still be a separate object as in 1), but have Core ban all known characters and IP addresses at ban time. This still suffers from some of the issues with 1), but removes the admin effort issue, and makes it so Core could use reference fields to track related bans. However, it still leads to some issues with logical layout and auto-banning as well as making bugs more likely with having to repopulate the reference field of every related ban when a new related ban is added.
3) Have a higher level object for each Ban action taken by an administrator (which from a logical, non-technical perspective would be considered 1 ban), and have that object contain the individual Character and IP Address Bans that resulted from that Ban action. This has several advantages over the prior 2 systems, such as allowing 1 character to be banned multiple times without iffy overlaps, allowing Core to more easily detect and auto-ban characters who have been previously banned, and keeping the Database schema in a way that is more consistent with the general use case. It is however, unintuitive and complicated, which are it's main downsides.

That said, I am obviously biased towards the system I picked, and I do recognize that it is more complicated, but I feel that it is worth it. If you have another solution, I would love to hear it though.

jboning:
Okay, that seems like a good line of though. Put some of the reasoning from 3) here into the docs on the Ban class :)

PR #251, opened 2014-06-08
Add Permissions Model.

Acidity:
This feature is a WIP.

Future PRs:
1) Add UI for App owners to specify permissions and provide descriptions.
2) Update Core to use permissions, including having a way to populate the perms DB with core permissions.
3) Add UI for viewing permissions, including users and groups that have those permissions.
4) Add UI for assigning permissions to groups and users.

### DIFF
--- a/brave/core/application/model.py
+++ b/brave/core/application/model.py
@@ -7,7 +7,7 @@ from mongoengine import Document, EmbeddedDocument, EmbeddedDocumentField, Strin
 
 from brave.core.util.signal import update_modified_timestamp
 from brave.core.application.signal import trigger_private_key_generation
-from brave.core.permission.model import Permission
+from brave.core.permission.model import Permission, WildcardPermission
 
 
 log = __import__('logging').getLogger(__name__)
--- a/brave/core/application/model.py
+++ b/brave/core/application/model.py
@@ -7,6 +7,7 @@ from mongoengine import Document, EmbeddedDocument, EmbeddedDocumentField, Strin
 
 from brave.core.util.signal import update_modified_timestamp
 from brave.core.application.signal import trigger_private_key_generation
+from brave.core.permission.model import Permission
 
 
 log = __import__('logging').getLogger(__name__)
@@ -75,6 +76,18 @@ class Application(Document):
     
     def __unicode__(self):
         return self.name
+        
+    def delete(self):
+        """Deletes the application and any permissions added by the application."""
+        perms = Permission.objects()
+    
+        for p in perms:
+            if p.application != self:
+                continue
+                
+            p.delete()
+        
+        super(Application, self).delete()
 
 
 class ApplicationGrant(Document):
--- a/brave/core/application/model.py
+++ b/brave/core/application/model.py
@@ -51,6 +51,9 @@ class Application(Document):
     site = URLField(db_field='s')
     contact = EmailField(db_field='c')
     
+    # This is the short name of the application, which is used for permissions. Must be lowercase.
+    short = StringField(db_field='p', unique=True, regex='[a-z]+')
+    
     key = EmbeddedDocumentField(ApplicationKeys, db_field='k', default=lambda: ApplicationKeys())
     
     mask = EmbeddedDocumentField(ApplicationMasks, db_field='m', default=lambda: ApplicationMasks())
--- a/brave/core/application/model.py
+++ b/brave/core/application/model.py
@@ -52,7 +52,7 @@ class Application(Document):
     contact = EmailField(db_field='c')
     
     # This is the short name of the application, which is used for permissions. Must be lowercase.
-    short = StringField(db_field='p', unique=True, regex='[a-z]+')
+    short = StringField(db_field='p', unique=True, regex='[a-z]+', required=True)
     
     key = EmbeddedDocumentField(ApplicationKeys, db_field='k', default=lambda: ApplicationKeys())
     
--- a/brave/core/character/model.py
+++ b/brave/core/character/model.py
@@ -8,6 +8,8 @@ from mongoengine import Document, StringField, DateTimeField, ReferenceField, In
 from brave.core.util.signal import update_modified_timestamp
 from brave.core.key.model import EVECredential
 from brave.core.util.eve import api
+from brave.core.permission.model import Permission, WildcardPermission, GRANT_WILDCARD
+from brave.core.application.model import Application
 
 
 log = __import__('logging').getLogger(__name__)
@@ -164,12 +166,13 @@ class EVECharacter(EVEEntity):
     
     titles = ListField(StringField(), db_field='ti', default=list)
     roles = ListField(StringField(), db_field='ro', default=list)
-    personal_permissions = ListField(ReferenceField('Permission', db_field='p'))
+    personal_permissions = ListField(ReferenceField(Permission), db_field='p', default=list)
     
     credentials = ListField(ReferenceField(EVECredential, reverse_delete_rule=PULL), db_field='e', default=list)
     
     owner = ReferenceField('User', db_field='o', reverse_delete_rule=NULLIFY)
     
+    # DEPRECATED
     @property
     def tags(self):
         from brave.core.group.model import Group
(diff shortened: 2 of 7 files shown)


Acidity:
This PR should be feature-complete, since I want to split this feature into multiple PRs.

brave/core/character/model.py
```diff
@@ -183,6 +184,63 @@ def titlesort(i):
...
         return OrderedDict((i, mapping[i]) for i in sorted(mapping.keys(), key=titlesort))
     
+    def permissions(self, app=None):
+        """Return all permissions that the character has that start with core or app.
+        
+        An app of None returns all of the character's permissions."""
+        
+        from brave.core.permission.model import Permission
+        from brave.core.application.model import Application
+        from brave.core.group.model import Group
```
jboning, line 209: Hrm. I know this is a pattern we use elsewhere, I don't like it very much though. Importing at the top level means that if there is a problem we discover it early, instead of whenever we first execute this code path.

brave/core/permission/model.py
```diff
@@ -0,0 +1,88 @@
...
+    name = StringField(db_field='n')
+    description = StringField(db_field='d')
+    
+    @property
+    def application(self):
+        """Returns the application that this permission is for."""
+        
+        from brave.core.application.model import Application
+        
+        app_short = self.name[:(self.name.index('.'))]
```
jboning: I probably would have written this as self.name.split('.')[0]. The way you used here will error if there is not a period, while the split way is tolerant and just returns the whole string; not sure which way is better here (it might not matter).

brave/core/permission/model.py
```diff
@@ -0,0 +1,88 @@
...
+        # Mongoengine has no way to find objects based on a regex (as far as I can tell at least...)
+        perms = set()
+        
+        num_wildcards = wild.count(GRANT_WILDCARD)
+
+        # Loops through all of the permissions, then loops through the segments of wild between the periods.
+        for perm in Permission.objects():
+            # Splits both the wildcard permission provided and the permission being checked.
+            wild_segments = wild.split(r'.')
+            perm_segments = perm.name.split(r'.')
```
jboning: You know that r'' does not make a difference here, right? (no need to remove it, just fyi)

brave/core/permission/model.py
```diff
@@ -0,0 +1,88 @@
...
+            
+            # Loops through each segment of the wildcard. 'core.example.*.test.*' would have segments of 'core', 
+            # 'example', '*', 'test', and '*' in that order.
+            for w_seg in wild_segments:
```
jboning: I recommend `for (w_seg, perm_seg) in zip(wild_segments, perm_segments)` (with a length check first)

brave/core/permission/model.py
```diff
@@ -0,0 +1,88 @@
...
+    def __repr__(self):
+        return "Permission('{0}')".format(self.name)
+            
+    @staticmethod
+    def wildcardPerms(wild):
+        """Returns all permissions which wild provides access to. wild should be a string."""
+        
+        from brave.core.application.model import Application
+        
+        # If the provided string doesn't contain any wildcards, returnan empty set.
```
jboning: Return an empty set, or match one permission exactly?

nitpick: "return an"

brave/core/permission/model.py
```diff
@@ -0,0 +1,88 @@
...
+        if GRANT_WILDCARD not in wild:
+            return set()
+        
+        # Mongoengine has no way to find objects based on a regex (as far as I can tell at least...)
```
jboning, line 79: It can do string prefix matching, right? (I think.) That could help reduce the number of comparisons we need to do.

On the other hand, there likely aren't going to be enough permissions for this to be terrible, so adding that now would probably be premature optimization.

brave/core/permission/model.py
```diff
@@ -0,0 +1,88 @@
...
+        perms = set()
+        
+        num_wildcards = wild.count(GRANT_WILDCARD)
```
jboning: did this get used anywhere?

brave/core/character/model.py
```diff
@@ -183,6 +184,63 @@ def titlesort(i):
...
+        for group in self.tags:
+            group = Group.objects(id=group).first()
+            for perm in group.permissions:
+                # Append all of the group's permissions when no app is specified.
+                if not app:
+                    permissions.add(perm)
+                    continue
+                
+                # Permissions are case-sensitive.
+                if perm.startswith('core') or perm.startswith(app):
```
jboning: Check should include the trailing ".", otherwise "forums" will get all of "forums2"'s permissions as well

brave/core/character/model.py
```diff
@@ -183,6 +184,63 @@ def titlesort(i):
...
+    def permissions(self, app=None):
+        """Return all permissions that the character has that start with core or app.
+        
+        An app of None returns all of the character's permissions."""
+        
+        from brave.core.permission.model import Permission
+        from brave.core.application.model import Application
+        from brave.core.group.model import Group
+        
+        # Use a set so we don''t need to worry about characters having a permission from multiple groups.
```
jboning: extra '

jboning:
I really like the way this is looking! The global "core." permission namespace seems handy. We'll need to enforce that no apps are named "core". Also I'd suggest automatically filling in "core.group.<foo>", to make it easy for apps that use the existing tags mechanism to switch over to the new permissions.

brave/core/character/model.py
```diff
@@ -183,6 +184,63 @@ def titlesort(i):
...
+        
+        return permissions
+        
+    def has_permission(self, permission):
+        """Accepts both Permission objects and Strings."""
+        
+        from brave.core.group.model import Permission
+        
+        if isinstance(permission, str) or isinstance(permission, unicode):
+            perm = permission
```
jboning: having both "perm" and "permission" is confusing. perm_string or permission_string would be better

brave/core/character/model.py
```diff
@@ -183,6 +184,63 @@ def titlesort(i):
...
+        
+        # Use a set so we don''t need to worry about characters having a permission from multiple groups.
+        permissions = set()
+        
+        # Allow app to be either the short name or the application object itself.
+        if isinstance(app, Application):
+            app = app.short_name
+        
+        # Return permissions from groups that this character has.
+        for group in self.tags:
```
jboning: Kinda weird to work with group objects, convert to strings, and then load all the groups from the database again. Maybe refactor the group membership logic into a groups method or property and then call it from this and tags?

brave/core/application/model.py
```diff
@@ -72,6 +76,18 @@ def __repr__(self):
     
     def __unicode__(self):
         return self.name
+        
+    def delete(self):
+        """Deletes the application and any permissions added by the application."""
```
jboning, line 81: Groups and characters will still have these permission strings lying about. Should consider whether that's okay (and if so, document so and why)

jboning:
Also it's probably pointing out the obvious to note that the wildcard thing isn't used yet, isn't it :P

brave/core/permission/model.py
Acidity: I'd prefer that it fail loudly rather than silently, since there should be no permissions without a dot (All permissions are required to start with {appName}.)

brave/core/permission/model.py
Acidity: Yes, but I prefer to use raw strings for regexes anyways.

brave/core/permission/model.py
jboning: sgtm

brave/core/permission/model.py
jboning: the argument to split() is just a plain old string though :)

Acidity: That it is... Too much Java programming I guess : P

brave/core/permission/model.py
Acidity, line 79: The problem with that is that although it might be faster (assuming MongoDB is more efficient than Python, which it probably is), it (in my opinion) significantly decreases the readability of the code as it implies that there is something 'special' about the first segment, and because it would require a special case for the permission string '*'.

brave/core/application/model.py
Acidity, line 81: Hmm, I'm not sure what the best way to go about this is... I'm leaning towards leaving them since it's much easier to delete extraneous permissions than it is to re-add an application's permissions to all the groups.

Acidity:
I'm not planning on removing the tags kw in Core.Info in this release (0.7), so I don't really feel that a pre-populated core.group permission is necessary.

brave/core/application/model.py
jboning, line 81: I'm nervous about someone deleting a permission and recreating it thinking that the old one was wiped clean. I'm trying to think of things we could do to mitigate this. Maybe warn when creating a permission that users/groups already have access to? Wildcard permissions might set that off all the time though. Hmmm.

Acidity, line 81: Yeah, I think a warning would be the best solution, because deleting the permissions setup of a 50+ group Core instance would be very bad.

jboning:
Re tags and groups, yeah, it's certainly not necessary to put group-permissions in right here. But we should definitely do it well before removing the tags keyword, for smooth transitioning of the apps.

Acidity:
Hmm, but the point of the permissions system is to prevent groups being used for application-side access control. Plus that would reveal all of the groups in Core, and who is in them (if they authorize the app), which is a security risk. I think that if we leave a decent amount of time between the deployment of 0.7 and when we remove 'tags', that should serve as the transition period rather than allowing applications to continue with bad practices indefinitely.

Acidity:
TODO: Fix has_permission to return true when passed a permission that does not exist but that a user has a wildcard that would still grant access for.

EDIT: Done.

jboning:
tests <3 <3 <3 <3 <3 <3

PR #253, opened 2014-06-09
Make the character portraits in the key table link to the appropriate...

Acidity:
...character detail pages.


jboning:
same as the other PR, kicking off a rebuild just to make sure.

PR #254, opened 2014-06-09
Rename the admin link in the side navigation bar to Admin Panel.

Acidity:


jboning:
I don't see how this could possibly have caused a build failure, I kicked off a rebuild so hopefully it turns out to be okay.

PR #255, opened 2014-06-09
Fix a bug that led to API key verification codes ebing displayed in their...

Acidity:
... entirety.


jboning:
I'm not sure how important this is. But I guess not leaking your keys if your core account is compromised seems admirable.

PR #256, opened 2014-06-09
Fix a bug where duplicate keys could be added to Core.

Acidity:


jboning:
This is probably going to need database cleanup to ensure no existing duplicate keys

Acidity:
I noticed this bug several weeks ago, and had an interim solution in place until I could identify the root cause, so no duplicates on the brave server at least.

PR #261, opened 2014-06-23
Add ability for applications to have different grant expiration times.

Acidity:


brave/core/application/controller/manage.py
```diff
@@ -51,7 +51,8 @@ def get(self):
                                 ),
                             required = app.mask.required,
                             optional = app.mask.optional,
-                            groups = app.groups
+                            groups = app.groups,
+                            expire = app.expireGrantTime
```
jboning: Python is fine with trailing commas, and it saves the extra -/+ in the diff when adding a new field. (Super minor, no need to change here, just a habit I like.)

brave/core/application/controller/manage.py
```diff
@@ -80,6 +81,9 @@ def post(self, **kw):
         app.mask.required = valid['required'] or 0
         app.mask.optional = valid['optional'] or 0
         
+        if user.admin:
```
jboning, line 84: Only core admins can change grant expiration?

brave/core/application/controller/manage.py
jboning: More substantially, expireGrantTime is unclear as a name, expireGrantDays would be better

brave/core/application/controller/manage.py
Acidity, line 84: Yes

PR #262, opened 2014-06-23
Fix bug that prevented the deletion of groups.

Acidity:

### DIFF
--- a/brave/core/group/template/list_groups.html
+++ b/brave/core/group/template/list_groups.html
@@ -46,7 +46,7 @@
                     },
                     function()
                     {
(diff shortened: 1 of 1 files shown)


mglinski:
I approve this message.
-Not Me

brave/core/group/template/list_groups.html
```diff
@@ -46,7 +46,7 @@
                     },
                     function()
                     {
-                        $.ajax('/group/' + row.attr('data-id') + '/delete').
+                        $.ajax('/group/' + row.attr('data-id') + '/delete', {type: 'POST'}).
```
ayust: A more concise version of this might be to change `$.ajax` to `$.post` instead.

PR #271, opened 2014-07-18
Implement Permissions checks into Core

Acidity:
I may have missed some spots...


brave/core/application/controller/manage.py
```diff
@@ -29,12 +32,17 @@ def __init__(self, app):
         except Application.DoesNotExist:
             raise HTTPNotFound()
 
-        if self.app.owner.id != user.id and not user.admin:
+        if self.app.owner.id != user.id and not user.has_permission('core.application.edit.'+str(app)):
```
jboning, line 35: What does app stringify to?

brave/core/application/controller/manage.py
```diff
@@ -80,6 +90,9 @@ def post(self, **kw):
         app.key.public = valid['key']['public']
         app.mask.required = valid['required'] or 0
         app.mask.optional = valid['optional'] or 0
+        app.short = valid['short'] or app.name.replace(" ", "").lower()
+        
+        createPerms(valid['perms'])
```
jboning: do we want to limit this to perms scoped to this app?

brave/core/application/controller/manage.py
```diff
@@ -145,6 +159,16 @@ def post(self, **kw):
             app.development = True
         else:
             app.development = False
+            
+        app.short = valid['short'] or app.name.replace(" ", "").lower()
```
jboning, line 186: App names can contain dots!

brave/core/application/controller/manage.py
```diff
@@ -145,6 +159,16 @@ def post(self, **kw):
...
+        
+        createPerms(valid['perms'])
+        
+        p = Permission(app.short+'.authorize', "Ability to authorize application {0}".format(app.name))
```
jboning: We probably want to verify that the app short is unique before creating the corresponding permissions

brave/core/application/form.py
```diff
@@ -30,6 +30,8 @@ def manage_form(action='/application/manage/'):
             Tab('perms', L_("Permissions"), children=[
                     TextField('required', L_("Required Mask"), placeholder='0', class_="input-small"),
                     TextField('optional', L_("Optional Mask"), placeholder='0', class_="input-small"),
+                    TextField('short', L_("Application Permission"), class_="input-block-level"),
```
jboning: Dunno if "Application Permission" is descriptive enough

brave/core/application/controller/manage.py
Acidity, line 35: It'll be the application's id (the one used by the API as an identifier, etc.), though I suppose it might be worth changing that to the application's short, that would break from the tradition of using id's for permissions (accounts, characters, groups, etc.).

jboning, line 35: The other places where we use IDs are database reference fields that get resolved into actual objects any time we use them, right? Whereas my impression is that permission strings are going to be used by humans at least a fair bit. If so, I think the app short is nicer.

brave/core/permission/util.py
```diff
@@ -0,0 +1,87 @@
...
+from brave.core.account.model import User
+from web.core.http import HTTPForbidden, HTTPUnauthorized
+from brave.core.util.predicate import authenticate
+import web.auth
+
+
+log = __import__('logging').getLogger(__name__)
+
+
+def user_has_permission(perm=None, wild=False, **runkw):
```
jboning: I have slight misgivings about adding wildcard matching here. Potentially a little bit confusing that we have permission-strings with stars in them granting many permissions under one usage, and accepting many strings in another.

renaming "wild" to something more descriptive here might help ("accept_any_matching" or something? low on good ideas)

brave/core/permission/util.py
```diff
@@ -0,0 +1,87 @@
...
+        @authenticate
+        def check_permission(self, *args, **kwargs):
+            user = web.auth.user
+            
+            # If there is no permission provided, then any auth'd user has permission
+            if not perm:
+                log.debug('No permission provided.')
+                return function(self, *args, **kwargs)
+            else:
+                permission = perm
```
jboning, line 37: permission/perm are confusing in this function

brave/core/permission/util.py
```diff
@@ -0,0 +1,87 @@
...
+                for attr in valSplit:
+                    if attr == 'self':
+                        value = self
+                        continue
+                    elif attr in kwargs:
+                        value = kwargs.get(attr)
+                        continue
+                    value = getattr(value, attr)
+                    
+                permission = perm.replace('{'+key+'}', value)
```
jboning: I don't understand the relationship between permission and perm in this function exactly, but this can't be what you meant :P

brave/core/permission/util.py
```diff
@@ -0,0 +1,87 @@
...
+            
+            # Cycle through the user's permissions, and if they have it leave the method.
+            for c in user.characters:
+                if c.has_permission(permission):
+                    return function(self, *args, **kwargs)
+            
+            # If wild=True, then we let any user with a permission granted by perm proceed.
+            if wild:
+                p = WildcardPermission.objects(id=permission)
+                if len(p):
```
jboning: what's this dance for?

brave/core/application/controller/manage.py
Acidity, line 35: Yes I know, but I meant that the permission for viewing a character's page is core.character.view.aswergsdfsd256456 as well, because the page is /character/sfwefsdfsfs23424, so would we want to change the permissions to use the human-friendly version for those? The issue if we do that is what is the human friendly version? For app's it's short, for character's it's name and for user's it's username? Not very consistent :P

brave/core/permission/util.py
Acidity, line 37: Python throws a fit if we try to modify perm directly, so to work around that I made a second variable (permission) to take perm's value and modify it as we needed.

brave/core/application/controller/manage.py
jboning, line 35: Hrm. I'm questioning the need for such fine-grained permissions in any context except apps. When would we want to grant someone permission to see exactly one user or one character?

brave/core/application/controller/browse.py
```diff
@@ -48,8 +48,18 @@ def index(self):
...
     @authenticate
     def index(self):
-        records = Application.objects(development__in=[False, None])  # TODO: This needs to be ACL-filtered!
-        devRecords = Application.objects(development=True)
+        records = list()
+        devRecords = list()
+        
+        # Retrieve non-development applications that the user is able to authorize
+        for r in Application.objects(development__in=[False, None]):
+            if user.has_permission(r.short+'.authorize'):
```
jboning: What are your thoughts on having this in the app's permission namespace vs the core namespace (like core.application.authorize.<app>)? My instinct would be to prefer the latter

Acidity: I initially had it be that latter, but changed it to application.authorize in order to more closely tie it to the application it's granting access to. I'm certainly open to reconsidering that though.

brave/core/application/controller/manage.py
```diff
@@ -29,12 +32,20 @@ def __init__(self, app):
...
-        if self.app.owner.id != user.id and not user.admin:
+        if self.app.owner.id != user.id and not user.has_permission('core.application.edit.'+str(app)):
             raise HTTPNotFound()
     
     def get(self):
         app = self.app
         
+        perms = ""
+        
+        for p in Permission.objects(id__startswith=app.short):
```
jboning: This is going to have the prefix problem (an app with short "app" will match permissions for "app." and "application."). Retrieving an application's permissions should be a model-level method (not sure whether on Permission or on Application)

brave/core/application/controller/manage.py
Acidity, line 35: Granting a spy-checker the ability to view information on a suspicious character/account

brave/core/application/controller/manage.py
```diff
@@ -80,6 +99,12 @@ def post(self, **kw):
         app.key.public = valid['key']['public']
         app.mask.required = valid['required'] or 0
         app.mask.optional = valid['optional'] or 0
+        app.short = valid['short'] or app.name.replace(" ", "").lower()
```
jboning: I'm still worried about mutating the app's short. This feels like it could lead to a lot of stale permissions in the DB, and permissions that people had going bad.

brave/core/group/controller.py
```diff
@@ -112,7 +142,16 @@ def post(self, id=None, title=None):
         if not g:
             return 'json:', dict(success=False,
                                  message=_("group with that id already existed"))
-
+        # Give the creator of the group the ability to edit it and delete it.
+        editPerm = Permission('core.group.edit.acl.'+g.id, "Ability to edit ACLs for Group {0}".format(g.id))
+        deletePerm = Permission('core.group.delete.'+g.id, "Ability to delete Group {0}".format(g.id))
+        # Might be called when we save the user... Should probably check that.
```
jboning: reminder? :)

brave/core/group/template/group.html
```diff
@@ -324,6 +347,30 @@
             <button class="addrule" data-rule="mask">by key mask</button>
         </div>
         <button id="submit">submit</button>
+        % endif
+        
+        <br>
```
jboning, line 351: I'm not an expert, but I don't think this is approved web layout technique :)

No need to change now, just a candidate for future cleanup

brave/core/public/js/script.js
```diff
@@ -208,68 +208,6 @@ $(function(){
         // $.post("url", {});
     }
     
-    // Generic field validation.
```
jboning, line 211: What happened here? Was this not being used at all?

jboning:
A thing I'd really like to have happen is for the various permissions be associated with their models, e.g. `Application.get_create_permission()` or `app.get_edit_permission()`. Or perhaps factored to actually contain the permissions checks, like `Application.user_can_create(user)` or `app.user_can_edit(user)`. Having the string construction scattered about is redundant and means that every time you want to check a permission you need to remember the structure of that permission (is it core. or app., does it use an id or a short or something else)

jboning:
... I'm not sure whether that should happen as part of this change, or further cleanup on this branch before merging it into master

Acidity:
I'd say that should be pushed off to another PR given "(3:45:41 AM) Arnath_Othorem: acid mate reviewing these things is way easier when they are smaller :V"

brave/core/public/js/script.js
Acidity, line 211: err, I didn't really intend to push that into the PR... but either way, it was what was causing the issue on /account/register (though it was being used by the application editing interface so, TODO: fix that in a future PR). An updated version of that code is available in /public/js/validation.js

brave/core/application/controller/manage.py
Acidity: Well as long as we don't delete permissions, the only issue is wasted space/people having pointless permission. While I agree that doing nothing isn't necessarily the best course of action, wiping an entire permissions set up would be awful as well. What we could do is take all of an app's permissions and rename them when the app short is changed however

brave/core/group/controller.py
Acidity: If your document contains ReferenceField or GenericReferenceField objects, then by default the save() method will not save any changes to those objects.

PR #275, opened 2014-07-29
Add permissions to the models of the objects they grant permissions for.

Acidity:

### DIFF
--- a/brave/core/account/model.py
+++ b/brave/core/account/model.py
@@ -47,7 +47,7 @@ class User(Document):
     other_accs_IP = ListField(ReferenceField('User'), db_field='otherAccountsIP')
     
     # Permissions
-    view_perm = 'core.account.view.{account_id}'
+    VIEW_PERM = 'core.account.view.{account_id}'
     
     # Python Magic Methods
     
@@ -225,9 +225,9 @@ class User(Document):
         other.save()
     
     @property
-    def get_view_perm(self):
+    def view_perm(self):
         """Returns the permission required to view this user's account details."""
-        return self.view_perm.replace("{account_id}", str(self.id))
+        return self.VIEW_PERM.replace("{account_id}", str(self.id))
 
 class LoginHistory(Document):
     meta = dict(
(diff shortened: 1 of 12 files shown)


brave/core/account/model.py
```diff
@@ -220,6 +223,11 @@ def remove_duplicate(acc, other, IP=False):
                 
         acc.save()
         other.save()
+    
+    @property
+    def get_view_perm(self):
```
jboning: style comment which applies throughout: usually I'd expect things starting with "get" to be functions. As a property I'd just name this `view_perm`

brave/core/application/model.py
```diff
@@ -65,6 +65,11 @@ class Application(Document):
     
     owner = ReferenceField('User', db_field='o')
     
+    # Permissions
+    edit_perm = 'core.application.edit.{app_short}'
+    create_perm = 'core.application.create.{app_short}'
```
jboning: don't think this is supposed to contain `.{app_short}` :)

brave/core/account/model.py
```diff
@@ -220,6 +223,11 @@ def remove_duplicate(acc, other, IP=False):
...
+        """Returns the permission required to view this user's account details."""
+        return self.view_perm.replace("{account_id}", str(self.id))
```
jboning: instead of replace, you should be able to say `.format(account_id=str(self.id))`, right? (Not that it makes a huge difference here.

brave/core/application/model.py
```diff
@@ -78,6 +83,17 @@ def __repr__(self):
     
     def __unicode__(self):
         return self.name
+    
+    def get_perm(self, perm_type):
+        return getattr(self, perm_type+"_perm").replace("{app_short}", self.short)
```
jboning: now this might be a little overabstracting :P

brave/core/account/model.py
Acidity: Should be able to, but I can never remember what format can and cannot do. I'll test it out

jboning:
OK I think this looks good

PR #276, opened 2014-07-29
Get rid of core.admin.search, and make the search page only show an...

Acidity:
...option to search based on the search types a user has permission for.


jboning:
The purist in me is a little bit sad that we have these permission strings duplicated, but I'm not sure where to put them exactly (the search controller is the best I can come up with?). But since they don't have anything substituted in them I think this way is okay.

PR #277, opened 2014-07-31
Add alternate script for updating keys.

Acidity:

### DIFF
--- a/brave/core/scripts/update_keys.py
+++ b/brave/core/scripts/update_keys.py
@@ -6,35 +6,30 @@ from datetime import datetime, timedelta
 from requests.exceptions import HTTPError
 
 class CredentialUpdateThread(Thread):
-    def __init__(self, indices, separation):
+    def __init__(self, key_groups, start_time, interval):
         super(CredentialUpdateThread, self).__init__()
-        self.indices = indices
-        self.separation = separation
-    
-    @property
-    def next_index(self):
-        for i in self.indices:
-            if i > UpdateKeys.current_index:
-                return i
-        return self.indices[0]
+        self.key_groups = key_groups
+        self.wake_time = start_time
+        self.interval = interval
+        self.group = 0
     
     def run(self):
         while True:
-            now = datetime.now()
-            index = self.next_index
-            next_index = now + timedelta(minutes=(self.next_index-UpdateKeys.current_index) if self.next_index > UpdateKeys.current_index else self.separation)
-            next_index = next_index - timedelta(microseconds=(next_index.microsecond + next_index.second*1000000))
-            sleep((next_index-datetime.now()).total_seconds())
-            # There must've been a delay somewhere, recalulate when it's this thread's turn to run
-            if not UpdateKeys.current_index + 1 == index and not UpdateKeys.current_index + 1 == index + self.separation:
-                continue
-            UpdateKeys.current_index = index
-            if UpdateKeys.current_index not in UpdateKeys.keys.keys():
-                continue
-            for k in UpdateKeys.keys[UpdateKeys.current_index]:
+            if datetime.now() > self.wake_time:
+                print "warning: no delay between groups! Likely falling behind requested time between pulls."
+            
+            # wait for the next wake time to arrive
+            while datetime.now() < self.wake_time:
+                sleep((self.wake_time - datetime.now()).total_seconds())
+            
+            # process the next batch of keys
+            for k in self.key_groups[self.group]:
                 self.update_key(k)
-        
-                    
+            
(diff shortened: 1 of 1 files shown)


brave/core/scripts/update_keys.py
```diff
@@ -0,0 +1,67 @@
...
+            key_result = key.pull()
+        except HTTPError as e:
+            print("Error {}: {}".format(e.response.status_code, e.response.text))
+        if not key_result:
+            print("Removed disabled key {0} from account {1} with characters {2}".format(k, key.owner, key.characters))
+
+def main(time_between_pulls=1440, threads=1):
+    """ Max supported int for time_between_pulls is 1440, setting time_between_pulls less than threads will probably
+        cause problems, so don't do it. This script makes no guarentee of thread safety, so using more than 1 thread
+        is strongly discouraged if you value your database's integrity."""
```
jboning: Even if you have only one thread going here, the core web site will be running concurrently.

jboning:
Can you talk a bit about why this is needed?

brave/core/scripts/update_keys.py
Acidity: Yeah I know, but not much we can do there :/

brave/core/scripts/update_keys.py
```diff
@@ -0,0 +1,67 @@
...
+        index = random.randint(0, time_between_pulls)
+        print "Assigning key {0} to {1}".format(k, index)
+        if index in keys.keys():
+            keys[index].append(k.key)
+        else:
+            keys[index] = [k.key]
+    
+    
+    for i in range(0, threads):
+        indices = []
```
jboning: `indices = range(i, time_between_pulls, threads)`?

brave/core/scripts/update_keys.py
```diff
@@ -0,0 +1,67 @@
...
+        if not key_result:
+            print("Removed disabled key {0} from account {1} with characters {2}".format(k, key.owner, key.characters))
+
+def main(time_between_pulls=1440, threads=1):
+    """ Max supported int for time_between_pulls is 1440, setting time_between_pulls less than threads will probably
+        cause problems, so don't do it. This script makes no guarentee of thread safety, so using more than 1 thread
+        is strongly discouraged if you value your database's integrity."""
+    # Seed the keys into a dictionary, sorted by randomly assigned integers
+    for k in EVECredential.objects():
+        index = random.randint(0, time_between_pulls)
```
jboning: any reason for random instead of just partitioning them equally? (of course in expectation they'll be about even, but seems nicer to have a guarantee)

Acidity:
As opposed to the script you wrote? Not much since the threading doesn't work. But it does do it's own timing, and spread key pulls out semi-evenly so it's got that as a bonus. As for why we want timed key pulls, ideally I'd like to remove the EVE API call when core.info is called so that we can have applications use that often to update against Core's permissions and (eventually) bans. Given that that call currently takes in the range of 2.5 to 3 seconds, that leads to significant issues with Mumble/Jabber logins.

jboning:
Also, confession, I don't really understand the logic works for determining how long to sleep. The 1440-max requirement seems a little strange--I suspect that an implementation could be simpler and not enforce this requirement.

jboning:
Keeping a long-running thread to do regular key pulls is nice, I like that better than telling everyone to set up a cron job. I think ideally we'd spawn off a thread as part of the application startup and not need a separate script. (But on the other hand if a core instance got big enough to need multiple servers, it would probably not want each of them running duplicate update jobs. hmmm.)

Acidity:
In a single thread instance, it will sleep until the next minute. On multithreaded instances, a thread sleeps until the next minute that was assigned to it (so for a 10 thread version with the thread's pulls taking 5 minutes 45 seconds, it would sleep for 4 minutes 15 seconds As for why the 1440 issue exists, I just realized that this works _only_ for 1440 because I forgot to test it for less than that... But the number used as the index is the minute of the day it's going to be pulled on.

Acidity:
As for why I _didn't_ have it in the init script was that exact reason, Brave's Core setup has 2 instances running next to each other, so putting it in the init script would lead to 2 of these running.

jboning:
In general I don't think I like this approach to timing the key pulls very much. There's no protection against exceeding CCP's rate limits (it just iterates through all the keys assigned to a particular timeslot). I'm not sure what other benefit there is to spreading the updates throughout a 24-hour (or other) period like this is.

jboning:
Cool, that makes sense as far as keeping this in its own file then!

Acidity:
Looks good \o/

PR #278, opened 2014-07-31
Allow applications to require a minimum key mask for authorization.

Acidity:


brave/core/util/eve.py
```diff
@@ -389,6 +389,9 @@ def __repr__(self):
     def has_access(self, mask):
         if isinstance(mask, EVEKeyMask):
             mask = mask.mask
+        # If the mask we're checking for is greater than this object's mask it has to be False
+        if mask > self.mask:
```
jboning: this is not how one compares bitvectors :P

jboning: Though, it might be worth adding comparison methods to the EVEKeyMask class so something like this would be correct

Acidity: I think you may have missed the part below this... This is just to fix a bug where if we had a key with mask A trying to see if it had permissions for mask B, A & B always returns true if B > A even though A does not give permissions for B.

jboning: Ahh. Well, it's still not going to be right, for example mask=0b101, self.mask=0b110.

Acidity: yeah, math is hard...

jboning: theeere we go :) go ahead and clean up this check again while we're here

brave/core/controller.py
```diff
@@ -56,14 +56,22 @@ def get(self, ar):
...
                 dict(success=False, message=_("This application requires that you have a character connected to your"
                                               " account. Please <a href=\"/key/\">add an API key</a> to your account."),
                      ar=ar))
-            return 'brave.core.template.authorize', dict(success=True, ar=ar, characters=characters, default=default)
+            chars = []
+            for c in characters:
+                if c.credential_for(ar.application.mask.required):
+                    chars.append(c)
+            if chars:
+                default = u.primary or characters[0]
```
jboning: Does this work out right if u.primary is not in characters?

PR #280, opened 2014-08-02
Report empty /authorize requests as 400s

jboning:


brave/core/controller.py
```diff
@@ -45,8 +45,11 @@ def ar(self, ar):
         except AuthenticationRequest.DoesNotExist:
             raise HTTPNotFound()
     
-    def get(self, ar):
+    def get(self, ar=None):
         from brave.core.application.model import ApplicationGrant
+
+        if ar is None:
```
Acidity, line 51: Is there a particular reason this isn't 

```
if not ar:
```

Should work fine either way, but if not is more consistent with the rest of the code.

jboning, line 51: I wanted to handle the particular case of a missing argument. If it's passed as a query arg with an empty string I dunno what happens

Acidity, line 51: Looking at the rest of the code, it looks like ar is supposed to be the id of the ApplicationRequest, so I'd think catching an empty string would actually be a good thing, but I'm not particularly concerned about this.

jboning, line 51: Yeah, that probably wouldn't hurt. Mainly here I wanted not to throw an exception when people visit the URL for no reason. It might be worthwhile to think about the UX here and maybe add an explanatory message

PR #284, opened 2014-08-03
Joinable Groups

Acidity:
TODO (in a future PR):
Description for groups
UI for creating GroupCategories
UI for adding a group to a GroupCategory

### DIFF
--- a/brave/core/account/model.py
+++ b/brave/core/account/model.py
@@ -185,7 +185,7 @@ class User(Document):
         else:
             p = WildcardPermission(id=permission)
         for permID in self.permissions:
-            if p.grantsPermission(permID.id):
+            if p.grants_permission(permID.id):
                 return True
                 
         return False
--- a/brave/core/account/model.py
+++ b/brave/core/account/model.py
@@ -164,7 +164,9 @@ class User(Document):
         log.debug('Checking if user has permission {0}'.format(permission))
         
         # Check the primary character first, and if they have the permission return them.
-        for p in self.primary.permissions():
+        primary_perms = self.primary.permissions() if self.primary else []
+        
+        for p in primary_perms:
             if p.grants_permission(permission):
                 return self.primary
                 
--- a/brave/core/account/model.py
+++ b/brave/core/account/model.py
@@ -176,6 +176,20 @@ class User(Document):
                 
         return None
         
+    def has_any_permission(self, permission):
+        """Returns true if the character has a permission that would be granted by permission."""
+        from brave.core.permission.model import WildcardPermission
+        p = WildcardPermission.objects(id=permission)
+        if len(p):
+            p = p.first()
+        else:
+            p = WildcardPermission(id=permission)
+        for permID in self.permissions:
+            if p.grantsPermission(permID.id):
+                return True
+                
+        return False
+        
     @staticmethod
     def add_duplicate(acc, other, IP=False):
         """Marks other as a duplicate account to this account.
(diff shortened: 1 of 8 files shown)


brave/core/group/controller.py
```diff
@@ -30,16 +31,58 @@ def __init__(self, id):
         except Group.DoesNotExist:
             raise HTTPNotFound()
 
+    @user_has_permission(Group.EDIT_REQUESTS_PERM, group_id='self.group.id')
```
jboning, line 35: `@post_only` on all of these

brave/core/group/controller.py
```diff
@@ -116,18 +164,98 @@ def delete(self):
...
         
         return 'brave.core.group.template.list_groups', dict(
             area='group',
             groups=visibleGroups,
+            joinableGroups=joinableGroups,
+            requestableGroups=requestableGroups,
+            categories=GroupCategory.objects(members__in=visibleGroups),
+        )
+        
+    def leave(self, group):
```
jboning, line 211: I'm surprised adding extra methods to an HTTPMethod subclass like this works

brave/core/group/model.py
```diff
@@ -53,12 +74,53 @@ def permissions(self):
                 perms.add(p)
                 
         return perms
+        
+    def add_join_member(self, character):
+        """Use this to prevent duplicates in the database"""
```
jboning: hahahahahah mongo. (Or any database, really, in this case.) There's no guarantee that someone else hasn't inserted this character since we read the group from the db. Probably the best thing to do is explicitly document that this is best-effort but not a guarantee and move on.

brave/core/group/template/group.html
```diff
@@ -352,7 +408,7 @@
         <br>
         <br>
         <br>
-        % if web.user.has_permission('core.group.edit.perms.'+group.id):
+        % if web.user.has_permission('core.group.edit.perms.'+group.id) and not rule_set:
```
jboning: perm string still hardcoded

brave/core/group/controller.py
```diff
@@ -30,16 +31,58 @@ def __init__(self, id):
...
+        if not c:
+            return 'json:', dict(success=False, message=_("Character with that name not found."))
+            
+        if not c in self.group.requests:
+            return 'json:', dict(success=False, message=_("Character with that name has no request to join this group."))
+        
+        log.info("Adding {0} to group {1} via REQUEST_ACCEPT approved by {2}".format(c.name, self.group.id, user.primary))
+        self.group.add_request_member(c)
+        self.group.requests.remove(c)
+        self.group.save()
```
jboning, line 47: should these return success?

Acidity, line 47: For the add_request_member stuff? What's the concern there, that adding (or removing) to the list will fail, or that you think we should indicate failure when the character has already been added?

jboning, line 47: Sorry, "these" as in accept_request() etc. The function returns failures with error messages on the checks above, I assume we'd want to return success to the client here. (It looks like the js is just reloading the page regardless right now, but presumably if there's an error we'll want to flash it or something. Not sure what we do elsewhere.)

brave/core/account/model.py
```diff
@@ -180,6 +180,20 @@ def has_permission(self, permission):
                 
         return None
         
+    def has_any_permission(self, permission):
+        """Returns true if the character has a permission that would be granted by permission."""
```
jboning, line 184: copypaste much? :)

brave/core/group/controller.py
```diff
@@ -135,6 +276,10 @@ def post(self, id=None, title=None):
         if not id:
             return 'json:', dict(success=False,
                                  message=_("id required"))
+                                 
+        if id == 'manange':
+            return 'json:', dict(success=False,
+                                 message=_("You cannot name a group 'manage'"))
```
jboning, line 282: wait, what?

jboning, line 282: oh, I see. ok I guess

jboning:
haven't read it all but it's a feature branch so whatever

PR #285, opened 2014-08-03
Clarify violation text

jboning:


Acidity:
I liked the old way of handling violations :/
Triumvirate will love this change though

jboning:
gotta find the right balance between quirkiness and usability :P

PR #286, opened 2014-08-03
Remove key verification code from title attr

jboning:
why was this even here? what value did it add?


Acidity:
It allowed users to verify that they entered their vCode correctly... but I just realized that if it's wrong Core will just delete their key anyways...

PR #292, opened 2014-08-13
Remove the character set restrictions on passwords. Resolves #291

### LINKED ISSUE #291: @ not permitted in passwords?
User report from a newbie in my alliance:

[03:05:14] Neven Cengari > it didn't like the @ in the password

Why are we restricting password keyspaces?

Acidity:


jboning:
This seems basically good to me, but I do worry a little bit that this will increase support workload (if there are people who think that whitespace is ignored or something crazy like that). Curious for thoughts from @TriumvirateRhade on that angle.

jboning:
based on jabber discussion it sounds like we're OK with this

PR #300, opened 2014-08-14
Feature/permissions fixes

Acidity:


brave/core/application/controller/manage.py
```diff
@@ -103,8 +103,10 @@ def post(self, **kw):
         
         if user.admin:
             app.expireGrantDays = valid['expire'] or 30
-        
-        if not createPerms(valid['perms'], app.short):
+            
+        app.short = valid['short'] or app.name.replace(" ", "").lower()
```
jboning, line 107: I still think we should make it hard to change an app's short

brave/core/character/template/charDetails.html
```diff
@@ -82,19 +82,33 @@ <h3 class="name">
                 </div>
             
                 <div class="permission-box" style ="padding-right: 25px;">
+                <%
```
jboning, line 85: this is too much logic for a template

PR #303, opened 2014-08-15
Add a notice on the log in page that the IGB is no longer supported if...

Acidity:
...we detect that the user is using the IGB.


jboning:
sadness though... someday I'd like to get this working again (not all of core but at least signin + app authorization)

PR #307, opened 2014-08-16
Groups UI Revamp and Interface Changes

mglinski:


jboning:
I confess that I haven't read this in great detail. Going to merge and then send a PR with a couple of tweaks

PR #309, opened 2014-08-16
tweaks to @mglinski's groups changes

jboning:


Acidity:
Note sure if user references in commit names emails the user so, @mglinski may want to take a look at this.

PR #311, opened 2014-08-17
Add support for key violations that describe insufficient API masks and ...

Acidity:
...incorrect API key types.


Acidity:
I'm not a huge fan of this solution to be completely honest, but I don't really want to redesign the EVECredential violation set up at the moment.

jboning:
Yeah, I'd like to rework this system a bit, but this seems like an okay stopgap.

PR #312, opened 2014-08-17
Fix a bug where characters were still able to have groups even if they...

Acidity:
...had no valid key on file.


Acidity:
Actually, I think this could be done better.

PR #315, opened 2014-08-17
Remove the API checks from api/core/info.

Acidity:
As there is now a script for automatically pulling API keys on a
regular basis, we can remove the API check from api/core/info,
substantially speeding up replies.


jboning:
https://www.youtube.com/watch?v=mk7VWcuVOf0&t=1m31s

PR #318, opened 2014-08-18
Cache user permissions on load for the group management page.

Acidity:

### DIFF
--- a/brave/core/group/controller.py
+++ b/brave/core/group/controller.py
@@ -263,8 +263,11 @@ class ManageGroupList(HTTPMethod):
         groups = sorted(Group.objects(), key=lambda g: g.id)
         
         visibleGroups = list()
+        
+        user_perms = user.permissions
+        
         for g in groups:
-            if user.has_permission(g.view_perm):
+            if Permission.set_grants_permission(user_perms, g.view_perm):
                 visibleGroups.append(g)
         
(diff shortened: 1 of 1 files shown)


jboning:
Hm, I don't particularly like this. Ideally the user model should encapsulate whatever caching it's doing.

Acidity:
Oh I agree, but this change has already been pushed into feature/permissions for a couple of other places as well and I noticed the significant decrease in performance when I updated so I thought I'd commit this until I had a better solution.

jboning:
file an issue so we don't forget about it and I'll merge =)

jboning:
Looking at this again, I'm not sure what I was thinking when I made that comment (tiredness?). I thought the controller was reaching into the user model, accessing some cache there, and using that. Just storing the set of permissions locally for use across several calls in a loop is totally fine.

(I think part of my confusion is the use of the term "caching" here. I can't come up with a good explanation why it would be an inappropriate term here. But it managed to throw me off.)

If this achieves good enough performance, I think doing it this way is better than the way in #320, since this avoids any nonobvious cache staleness. That kind of caching could still be useful if we find ourselves needing to make a lot of checks in a way where we can't just grab the set of permissions once and use it across all of the checks. But I'm inclined to avoid doing that until we have a need.

I'm really sorry for the fakeout here :(

PR #320, opened 2014-08-19
Add caching of permissions to the Character and User models. Resolves #319

### LINKED ISSUE #319: Improve performance of multiple permissions checks in one page load.
I.e., do caching in the character and user models.

Acidity:
This ensures that the permissions of the affected objects are evaluated
only once per object initialization (with the exception of app specific
permissions checks for EVECharacters). Prior to this change, every time
a user or character's permissions were checked, such as with
has_permission, the user's permissions were re-evaluated, which led to
substantial slow down on pages with a large number of permissions
checks, such as the group management page, (from feature/groups), the
application browse page, and the application management page.

### DIFF
--- a/brave/core/account/model.py
+++ b/brave/core/account/model.py
@@ -53,6 +53,11 @@ class User(Document):
     
     # Python Magic Methods
     
+    def __init__(self, *args, **kw):
+        super(User, self).__init__(*args, **kw)
+        
+        self._permissions = []
+    
     def __repr__(self):
         return 'User({0}, {1})'.format(self.id, self.username).encode('ascii', 'backslashreplace')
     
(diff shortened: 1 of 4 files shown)


brave/core/account/model.py
```diff
@@ -146,7 +161,7 @@ def delete(self):
         super(User, self).delete()
     
     @property
-    def permissions(self):
+    def __permissions__(self):
```
jboning, line 164: It turns out you're not supposed to do this with the **double_underscores**: "Never invent such names; only use them as documented" -- http://legacy.python.org/dev/peps/pep-0008/#descriptive-naming-styles

I'd probably just make this a normal method called `_load_permissions()` or something.

(Github was erroring this morning so I mentioned this on jabber but it probably got dropped on the floor among other conversation)

brave/core/account/model.py
```diff
@@ -66,6 +71,16 @@ def created(self):
     # Related Data Lookups
     
     @property
+    def permissions(self):
+        """Returns a user's permissions. This value is cached for the lifetime of the object, so actions which modify
+        a user's permissions will not be reflected in this value until the object is reinitialized. I don't think that
```
jboning, line 76: I guess this works for now, but I'm not terribly excited about this. It feels like something that would be very nonobvious if someone does end up tripping over it.

Acidity:
Made sort of unnecessary by #334.

PR #321, opened 2014-08-19
Added "Very Secure" ACL rule indicating OTP requirement.

amcgregor:
Untested, but the general idea is there.  Also a note about how "human_readable_repr" is doing it wrong.  Might submit a swap-out of that for the standard mechanism in a different merge request.

### DIFF
--- a/brave/core/group/acl.py
+++ b/brave/core/group/acl.py
@@ -210,3 +210,25 @@ class ACLMask(ACLRule):
                 not_=' not' if self.inverse else '',
                 mask=self.mask,
         )
+
+
+class ACLVerySecure(ACLRule):
+    """Grant or deny access based on mandatory use of an OTP."""
+    
+    def evaluate(self, user, character):
+        if user.rotp and len(user.otp):
+            return None if self.inverse else self.grant
+        
+        return self.grant if self.inverse else None
+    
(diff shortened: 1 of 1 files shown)


brave/core/group/acl.py
```diff
@@ -210,3 +210,25 @@ def human_readable_repr(self):
                 not_=' not' if self.inverse else '',
                 mask=self.mask,
         )
+
+
+class ACLVerySecure(ACLRule):
+    """Grant or deny access based on mandatory use of an OTP."""
+    
+    def evaluate(self, user, character):
+        if user.rotp and len(user.otp):
```
jboning, line 219: A similar expression is used elsewhere, I'll combine those into a method on the user model

brave/core/group/acl.py
```diff
@@ -210,3 +210,25 @@ def human_readable_repr(self):
...
+            return None if self.inverse else self.grant
+        
+        return self.grant if self.inverse else None
+    
+    def __repr__(self):
+        return "ACLVerySecure({0})".format(self.human_readable_repr())
+    
+    def human_readable_repr(self):
```
jboning, line 227: You can add this change to your other PR :)

PR #324, opened 2014-08-20
Added password strength meter to signup page.

seandsanders:


brave/core/account/template/signup.html
```diff
@@ -35,6 +36,9 @@
                 container: '#register-form'
                 }).popover('show');
         }
+
+        var ZXCVBN_RESULT_STRINGS = ['Fail', 'Really Shit', 'Shit', 'Fine, I Guess', 'Impregnable!'],
+        MINIMUM_PASSWORD_SCORE = 3;
```
jboning, line 41: ideally this value can come from where it's defined on the python side so we're not in danger of diverging

brave/core/account/template/signup.html
```diff
@@ -124,6 +128,20 @@
                 else {
                     self.popover('destroy');
                 }
+
+                var zxcvbnResult = zxcvbn(self.val());
+                $(".password-verdict").text(ZXCVBN_RESULT_STRINGS[zxcvbnResult.score]);
+                $(".password-meter .bar").width( (zxcvbnResult.score + 1) * 20 + "%");
```
jboning, line 134: Making the meter a smooth representation of entropy (rather than the not-very-granular score) might be nice, I think I will play with that.

brave/core/account/template/signup.html
```diff
@@ -124,6 +128,20 @@
...
+                if( zxcvbnResult.score < MINIMUM_PASSWORD_SCORE )
+                {
```
jboning, line 136: The rest of the project uses a denser brace style, it's nice to be consistent.

jboning:
This is a spectacular usability improvement. Thank you.

brave/core/account/template/signup.html
jboning, line 136: Hah, apparently that was a lie. Retracted.

PR #330, opened 2014-09-01
Add support for time based OTPs for two factor authentication.

Acidity:
Hopefully Google won't disable QR Code support in their Charts API any time soon...

### DIFF
--- a/brave/core/account/controller.py
+++ b/brave/core/account/controller.py
@@ -45,6 +45,15 @@ def _check_password(passwd1, passwd2):
 
 
 class TFA(HTTPMethod):
+    
+    @staticmethod
+    def clear_session():
+        """This nullifies all TFA related session values for the current session."""
+        session['auth'] = None
+        session['preauth_username'] = None
+        session['redirect'] = None
+        session.save()
+        
     def get(self):
         form = tfa_form()
         return 'brave.core.account.template.tfa', dict(form=form)
@@ -56,20 +65,16 @@ class TFA(HTTPMethod):
         user = User.objects.get(username=session['preauth_username'])
         
         if datetime.now() - session['auth'] > timedelta(minutes=15):
-            session['auth'] = None
-            session['preauth_username'] = None
-            session['redirect'] = None
-            session.save()
+            self.clear_session()
             return 'json:', dict(success=False, mmessage=_("Current Session has expired, please log in again"))
         
+        # TODO: Give them 3 tries to fail the OTP before requiring username and pass again
         if not user.otp.verify(OTP):
+            self.clear_session()
             return 'json:', dict(success=False, message=_("Incorrect One Time Password"))
         
         authenticate(user)
-        session['auth'] = None
-        session['preauth_username'] = None
-        session['redirect'] = None
-        session.save()
+        self.clear_session()
         
         return 'json:', dict(success=True, location=session['redirect'] or '/')
 
--- a/brave/core/account/controller.py
+++ b/brave/core/account/controller.py
@@ -74,7 +74,6 @@ class TFA(HTTPMethod):
             return 'json:', dict(success=False, message=_("Incorrect One Time Password"))
         
         authenticate(user)
-        self.clear_session()
         
         return 'json:', dict(success=True, location=session['redirect'] or '/')
 
--- a/brave/core/account/controller.py
+++ b/brave/core/account/controller.py
@@ -3,15 +3,16 @@
 from marrow.util.bunch import Bunch
 from marrow.mailer.validator import EmailValidator
 from web.auth import authenticate, deauthenticate, user
-from web.core import Controller, HTTPMethod, request, config
+from web.core import Controller, HTTPMethod, request, config, session
 from web.core.http import HTTPFound, HTTPSeeOther, HTTPForbidden, HTTPNotFound
 from web.core.locale import _
 from mongoengine import ValidationError, NotUniqueError
+from datetime import timedelta, datetime
 
 from brave.core.account.model import User, PasswordRecovery, TimeOTP, YubicoOTP
 from brave.core.account.form import authenticate as authenticate_form, register as register_form, \
-    recover as recover_form, reset_password as reset_password_form
-from brave.core.account.authentication import lookup_email, send_recover_email
+    recover as recover_form, reset_password as reset_password_form, tfa as tfa_form
+from brave.core.account.authentication import lookup_email, send_recover_email, verify_credentials
 from brave.core.util.predicate import is_administrator
 
 from yubico import yubico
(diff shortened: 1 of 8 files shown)


jboning:
An idea on the structure of login code: have the credential verification methods (password/2fa/whatever) only return a user object if the user is logged all the way in, and throw exceptions otherwise (with different exceptions for "password failed" and "needs 2fa" and whatever else). That should make it harder to accidentally mess up the controller and log someone in when they shouldn't be. (No need to actually refactor this way, but, this is going to need changes to solve merge conflicts, so if you feel it's worthwhile)

brave/core/account/model.py
```diff
@@ -2,17 +2,129 @@
...
+        field is a StringField across all subclasses, and as far as I know, it is not possible to add additional
+        restrictions to a field that is declared in a parent class. Note that the identifier in this case is the
+        shared secret."""
+        
+        if len(identifier) != 16:
+            raise ValidationError('The identifier for TimeOTPs must be 16 characters.')
+            
+        super(TimeOTP, self).__init__(identifier=identifier, *args, **kwargs)
+        
+    def verify(self, response):
```
jboning, line 54: A thing we should do as part of this is record that this OTP has already been used to log in, and refuse it in the future. (If you can use it more than once, it's not very one-time.)

brave/core/account/template/settings.html
```diff
@@ -256,6 +280,21 @@
...
+            % if web.user.otp:
+            
+            $('.viewotp').on('click', function()
+            {
+                display(
+                    "View OTP QR Code",
+                    '<p>' +
+                    'Scan the QR Code below to add the OTP.' +
+                    '</p>' +
+                    '<img src="https://chart.googleapis.com/chart?cht=qr&chs=200x200&chld=Q&chl=${web.user.otp.uri}">');
```
jboning, line 292: Normally I'd complain about sending key material to a 3rd-party service. I think we might be small enough fish that I'm not too terribly worried about it though.

brave/core/account/template/settings.html
```diff
@@ -644,7 +745,12 @@
             <a class="btn change"><i class="fa fa-cog"></i> Change Password</a><br /><br />
             <a class="btn changeemail"><i class="fa fa-cog"></i> Change Email Address</a><br /><br />
             <a class="btn configureotp"><i class="fa fa-cog"></i> Configure OTP</a><br /><br />
-            <a class="btn addotp"><i class="fa fa-cog"></i> Add OTP</a><br /><br />
+            % if not web.user.otp:
+            <a class="btn addyubicootp"><i class="fa fa-cog"></i> Add Yubico OTP</a><br /><br />
+            <a class="btn addtimeotp"><i class="fa fa-cog"></i> Add Time OTP</a><br /><br />
+            % else:
+            <a class="btn viewotp"><i class="fa fa-cog"></i> View OTP QR Code</a><br /><br />
```
jboning, line 757: Does this apply to yubikeys?

setup.py
```diff
@@ -43,6 +43,7 @@
                 'futures',
                 'zxcvbn',
                 'flake8',
+                'pyotp'
```
jboning: nit: never not trailing comma :P

jboning:
Other high-level thought is that it's nice to force people to enter a 2fa code as part of the activation flow, so it's harder for them to get locked out.

jboning:
Oh, other other high-level thought is, do we need a migration for the user model change from rotp/otp?

Acidity:
Regarding migration, at the moment there's very few people with yubikeys enabled on their Core accounts in brave core (since the yubikey functionality has been broken for months) so I'm just planning on setting those to None.

brave/core/account/controller.py
```diff
@@ -41,6 +42,57 @@ def _check_password(passwd1, passwd2):
...
+            user = User.objects.get(username=session['preauth_username'])
+        except User.DoesNotExist:
+            self.clear_session()
+            return 'json:', dict(success=False, message=_("Current Session has expired, please log in again"))
+        
+        if datetime.now() - session['auth'] > timedelta(minutes=15):
+            self.clear_session()
+            return 'json:', dict(success=False, message=_("Current Session has expired, please log in again"))
+
+        previously_used = OTPHistory.objects(otp=OTP).first()
```
jboning: include the user in the query here, rather checking on the next line, in case both this user and another user happen to have had the same OTP code

brave/core/account/model.py
```diff
@@ -238,3 +354,18 @@ def __repr__(self):
...
+class OTPHistory(Document):
+    meta = dict(
+        allow_inheritance=False,
+        indexes=['user', 'otp', dict(fields=['expires'], expireAfterSeconds=0)]
+    )
+
+    user = ReferenceField(User, required=True, db_field='u')
+    otp = StringField(db_field='o', required=True)
+    # Expire after 90 seconds, (support for at most a radius of 2 when checking OTP validity)
+    expires = DateTimeField(db_field='e', default=lambda: datetime.utcnow() + timedelta(seconds=90))
```
jboning, line 424: expires doesn't seem to be used ever?

Acidity, line 424: The deletion is handled automatically by mongoengine as specified in the indexes. http://docs.mongoengine.org/guide/defining-documents.html#indexes This is used in the API data caching as well.

jboning, line 424: Aha! Cute.

brave/core/account/model.py
```diff
@@ -33,9 +142,7 @@ class User(Document):
     confirmed = DateTimeField(db_field='c')
     admin = BooleanField(db_field='d', default=False)
     
-    # TODO: Extract into a sub-document and re-name the DB fields.
-    rotp = BooleanField(default=False)
-    otp = ListField(StringField(max_length=12), default=list)
+    otp = EmbeddedDocumentField(OTP, db_field='o')
```
jboning, line 145: It turns out this _does_ need a migration. Old accounts will have empty lists as their OTPs, which does not play nicely with the plain EmbeddedDocumentField validation.

Besides which, why do we want to remove the ListField? For the yubikey case, it's possible that someone would have two yubikeys which they just leave in different computers (esp. the nano: https://en.wikipedia.org/wiki/YubiKey#mediaviewer/File:YubiKey_Neo_and_Nano.jpg). We probably do want to restrict the user to only one TOTP, since having more than one makes it easier to brute force (and the search space is pretty small). Feels like this might be an argument for treating yubikey and totp differently.

jboning:
Since pyotp is pretty unsupported, what would you think about using https://cryptography.io/en/latest/hazmat/primitives/twofactor/#cryptography.hazmat.primitives.twofactor.totp.TOTP for this instead?

brave/core/account/template/settings.html
cheald, line 292: @jboning This could be replaced with something like qrcode.js to generate the QR code on the client, too, which would eliminate the third-party concern, and it's pretty much a drop-in replacement for this.

PR #332, opened 2014-09-01
Fix a bug with characters who had all numeric names.

Acidity:


jboning:
Ideally we'd have the API layer actually understand the various calls and the structure of their responses, but making it work seems okay for now

PR #333, opened 2014-09-02
Refactor the Permission checking decorators

Acidity:


brave/core/permission/util.py
```diff
@@ -11,39 +11,18 @@
...
 
 
-def user_has_permission(perm=None, accept_any_matching=False, **runkw):
-    """accept_any_matching=True means that any permission that perm would grant will return true.
-        @user_has_permission('core.group.edit.*', accept_any_matching=True) would work for any user with the ability to
-        edit anything about any group (for instance, a user with core.group.edit.acl.bob would be granted access when
-        accept_any_matching=True, but not when wild=False."""
-        
+def prepare_runtime_permission(perm=None, **runkw):
+    """This decorator handles the runtime permission aspects of the user permission checking decorators."""
```
jboning, line 15: I was pretty confused at first about what this actually did. Is there a reason you wrote it as a decorator rather than a normal helper function?

jboning:
I like the user_has_permission/user_has_any_permission split!

brave/core/account/model.py
```diff
@@ -193,6 +193,8 @@ def has_any_permission(self, permission):
         for permID in self.permissions:
             if p.grants_permission(permID.id):
                 return True
+            if permID.grants_permission(p.id):
+                return True
```
jboning, line 197: What's going on here?

brave/core/permission/util.py
```diff
@@ -11,27 +11,47 @@
...
-            if not perm:
-                log.debug('No permission provided.')
-                return function(self, *args, **kwargs)
-            else:
-                # Can't modify perm due to scoping, so we duplicate it's value into a 'local' variable and use that.
-                permission = perm
         
             user = user._current_obj()
+
+            permission = prepare_runtime_permission(self, perm, runkw, args, kwargs)
```
jboning, line 54: check for False here?

brave/core/permission/util.py
```diff
@@ -42,34 +62,53 @@ def check_permission(self, *args, **kwargs):
...
+        
+    def decorator(function):
+        
+        @authenticate
+        def check_permission(self, *args, **kwargs):
+            user = web.auth.user
+        
+            user = user._current_obj()
+
+            permission = prepare_runtime_permission(self, perm, runkw, args, kwargs)
```
jboning, line 96: same question

brave/core/permission/util.py
```diff
@@ -42,34 +62,53 @@ def check_permission(self, *args, **kwargs):
...
+                log.debug('User has no characters.')
+                raise HTTPForbidden()
+
+            p = WildcardPermission.objects(id=permission).first()
+            if not p:
+                p = WildcardPermission(id=permission)
+            for permID in user.permissions:
+                if p.grants_permission(permID.id):
+                    return function(self, *args, **kwargs)
+                if permID.grants_permission(p.id):
```
jboning, line 118: same question as the first one, this seems to be a behaviour change?

brave/core/account/model.py
Acidity, line 197: This fixes a couple of bugs, mainly centered around the fact that if we're checking for any permission of the form 'core.group.*', then the user having 'core.*' should pass evaluation.

PR #334, opened 2014-09-05
Improve performance of permissions related methods and functions.

Acidity:


jboning:
Is the list()/sort() stuff still needed?

I assume the set_has_permissions thing, and not expanding all the wildcards, doesn't actually change behaviour?

Acidity:
The list sort stuff isn't necessary, but I figure it might be nice to return permissions sorted anyways, especially since in my tests was very fast anyways.

The set_has_permissions change shouldn't affect anything functionally, and the lack of expanding wildcards should not lead to issues with permission evaluation.

brave/core/account/model.py
```diff
@@ -168,17 +170,7 @@ def has_permission(self, permission):
...
-            for p in self.primary.permissions():
-                if p.grants_permission(permission):
-                    return self.primary
-                
-        # Primary didn't have permission, check if the other characters do.
-        for c in self.characters:
-            for p in c.permissions():
-                if p.grants_permission(permission):
-                    return c
+        Permission.set_has_permission(self.permissions, permission)
```
jboning: it looks to me like this should be Permission.set_grants_permission(), and also have its return value checked. I'd really like to have enough automated testing to cover this sort of thing

PR #335, opened 2014-09-07
Fix a bug removing alliance from characters whose corp left an alliance.

Acidity:
Thanks Aura!


jboning:
ohai IG

PR #337, opened 2014-09-15
Remove sign in when authenticating already authorized apps.

Acidity:
I'm submitting this to start a discussion on whether it'd be worthwhile and to raise any security concerns that it might cause. In particular, given that there have been reports of session issues with the forums, I want to make 100% certain that there would be no such issues with Core. Additionally, it might be worth making sessions expire when someone tries to use that session id from a different IP address, though this will cause shortened session times for those with dynamic IP addresses, or that switch the location of their computer often.

### DIFF
--- a/brave/core/controller.py
+++ b/brave/core/controller.py
@@ -93,8 +93,14 @@ class AuthorizeHandler(HTTPMethod):
                      ar=ar))
                      
             return 'brave.core.template.authorize', dict(success=True, ar=ar, characters=chars, default=default)
-        
-        ngrant = ApplicationGrant(user=u, application=ar.application, mask=grant.mask, expires=datetime.utcnow() + timedelta(days=ar.application.expireGrantDays), character=grant.character)
+
+        # User had to log in for this authorization, so we extend the expiry.
+        if session.get('ar', None) == ar.id:
+            expiration = datetime.utcnow() + timedelta(days=ar.application.expireGrantDays)
+        else:
+            expiration = grant.expires
+
+        ngrant = ApplicationGrant(user=u, application=ar.application, mask=grant.mask, expires=expiration, character=grant.character)
         ngrant.save()
         
         ar.user = u
--- a/brave/core/controller.py
+++ b/brave/core/controller.py
@@ -36,9 +36,19 @@ class DeveloperTools(Controller):
 class AuthorizeHandler(HTTPMethod):
     def ar(self, ar):
         if not session.get('ar', None) == ar:
-            session['ar'] = ar
-            session.save()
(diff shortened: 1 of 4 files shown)


jboning:
I am in favor of doing this. Definitely worth thinking carefully about before we make the change though

brave/core/controller.py
```diff
@@ -35,10 +35,17 @@ def test(self):
...
+            try:
+                ar = AuthenticationRequest.objects.get(id=ar, user=None, grant=None)
+                grants = user.grants if user else []
+                for a in grants:
+                    if a.application == ar.application:
+                        return ar
+
+                raise HTTPFound(location='/account/authenticate?redirect=%2Fauthorize%2F{0}'.format(ar.id))
+            except AuthenticationRequest.DoesNotExist:
+                raise HTTPFound(location='/account/authenticate?redirect=%2Fauthorize%2F{0}'.format(ar.id))
```
jboning, line 48: `ar` won't be an object at this point

brave/core/controller.py
```diff
@@ -154,6 +166,20 @@ class RootController(StartupMixIn, Controller):
     admin = util.load('admin')
 
     def __call__(self, req):
+
+        if session and ('host' not in session or session['host'] != request.remote_addr):
```
jboning, line 170: Hrm, remote addr is not super reliable, people with laptops move around and home dhcp leases change. Would it be okay to let people change IPs?

brave/core/controller.py
```diff
@@ -154,6 +166,20 @@ class RootController(StartupMixIn, Controller):
...
+
+        session['host'] = request.remote_addr
+        session.save()
+
+
+        if user:
+            authorized_apps = ', '.join([str(g.application.id) for g in user.grants])
+            if request.cookies.get('authorized_apps') != authorized_apps:
+                sso_domain = config['core.sso.domain'] if config['core.sso.domain'] != 'localhost' else None
+                response.set_cookie('authorized_apps', authorized_apps, domain=sso_domain, overwrite=True)
```
jboning, line 181: What's the goal of this, again? It was so applications can message to the user that they'll need to log in?

Acidity, line 181: Yeah, right now if we just stick to the original commit, the user still needs to be prompted for and press the authorize/login button on the app. This cookie allows for Core to signal to the apps that the user has this app authorized, so they can automatically conduct the login flow so that applications automatically log you in if you have a valid Core account.

jboning:
Merged two commits from this. I'm downright rejecting the session IP binding, a shared cookie is probably okay but I want to think about it as a separate PR :)

PR #340, opened 2014-09-19
Comment out unused values

tyler274:
the api.\* values are used in applications but not core itself


Acidity:
Might be better to just remove the values all together.

tyler274:
Good point, I commented them out if only add some self documentation regarding it's function though it may as well not be there.

PR #347, opened 2014-10-16
Refactor the user tracking mechanism.

Acidity:


Acidity:
Note: Not complete nor ready for merging.

Acidity:
Think this should be good to go now (unless I forgot something...), obviously it's open to having additional means of identifying duplicate accounts in the future, but I'd like to get it merged before it gets too complex.

PR #350, opened 2014-10-30
Add support for multi-character grants.

Acidity:


jboning:
It seems weird that "all characters" is a bit independent of each individual character; it feels like I'd authorize either specific characters, or specific characters, not both. (I guess the advantage of doing it this way is that you can toggle all characters on and off again without losing information; but it still might be a bit confusing? I'm not sure which way is better.)

brave/core/api/core.py
```diff
@@ -182,22 +182,61 @@ def info(self, token):
...
+        for char in token.characters:
+            # Ensure that this character still belongs to this user
+            if char.owner != token.user:
+                token.remove_character(char)
+                try:
+                    token.reload()
+                except ApplicationGrant.DoesNotExist:
+                    break
+
+            # Fill in the character info
```
jboning: 202-228 and 233-242 are basically the same thing twice, could probably refactor a bit for less redundancy

brave/core/character/model.py
```diff
@@ -332,16 +332,18 @@ def detach(self):
...
-            self.owner.primary = None
-            self.owner.save()
-                    
-        # Delete any application grants associated with the character.
-        for grant in self.owner.grants:
-            if self == grant.character:
-                grant.delete()
-        
-        self.owner = None
+        if self.owner:
```
jboning, line 328: This change kind of deserved to be separate :)

PR #353, opened 2014-11-10
Groups Updates.

Acidity:


brave/core/group/controller.py
```diff
@@ -114,20 +181,117 @@ def delete(self):
...
+    def post(self, id=None, action=None):
+        if not action:
+            return 'json:', dict(success=False)
+        else:
+            group = Group.objects(id=id).first()
+            
+            if not group:
+                return 'json:', dict(success=False, message=_("Group not found"))
+            
+            return getattr(self, action)(group)
```
jboning: This is kind of scary, action could be _anything_ :P I'll send a PR to this branch to restrict to the things that we want.

PR #357, opened 2014-11-22
Make the update script more resiliant to errors.

Acidity:

### DIFF
--- a/brave/core/scripts/update_keys.py
+++ b/brave/core/scripts/update_keys.py
@@ -45,11 +45,18 @@ class CredentialUpdateThread(Thread):
     @staticmethod
     def update_key(k):
         key = EVECredential.objects(key=k).first()
+        if not key:
+            print "Key {} not found".format(k)
+            return
+        key_result = None
         try:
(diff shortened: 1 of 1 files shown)


brave/core/scripts/update_keys.py
```diff
@@ -45,11 +45,18 @@ def run(self):
...
+        if not key:
+            print "Key {} not found".format(k)
+            return
+        key_result = None
         try:
             print "Pulling key ID {0}".format(k)
             key_result = key.pull()
         except HTTPError as e:
             print("Error {}: {}".format(e.response.status_code, e.response.text))
+        except Exception as ex:
```
jboning, line 57: I'm really not convinced that any exception should trigger key removal. We definitely don't want to wipe everyone's keys over a typo making its way into pull() (though of course that should be impossible, right, with our test coverage, oh wait)

jboning:
Hm. The False/None thing is a little subtle, it feels like it would be easy to mess up later. Not sure what I would do instead

Issue #370, opened 2014-11-30
Don't drop keys on transient EVE API failure


ayust:
Right now occasionally keys will just get dropped due to a transient failure from the EVE API. It's really annoying to keep re-adding them.

The dropping occurs here:
https://github.com/bravecollective/core/blob/develop/brave/core/key/model.py#L229

Possible improvements I see:
1. Don't just check for a 403 HTTP error - actually check to see if there was an XML error returned by the API; for actual key issues, you'll get a "Authentication failure." with API error code 203. (It's possible that the transient errors also manifest themselves this way, but they might not. Worth properly handling this.)
2. Require multiple attempts to fail before actually deleting the key, to work around transient issues.

ayust:
Update: apparently the recurring failures return API error code 221 ("Illegal page request! Please verify the access granted by the key you are using!"). So we should be able to filter on that.

https://forums.eveonline.com/default.aspx?g=posts&m=4790500#post4790500

PR #371, opened 2014-11-30
Work around key drops from transient API failures

ayust:
Band-aid for #370


ayust:
This patch isn't perfect from a technical standpoint, but it has the virtue of being simple.

jboning:
I just pushed https://github.com/eve-val/core/commit/c91d2936095756cdff7252fde0ae67a24aa9ef66 to sound's deployment instead. No actual behaviour change, just logging.

My mention of error code 221 earlier was probably misleading--that's the error we've been seeing on posmon, no idea if it's the same here. I'm guessing it's likely, but it's corp vs character keys, so who knows.

PR #376, opened 2014-12-06
Fix the proxy API for char calls.

Acidity:


jboning:
Can we make this default to the grant's primary character if no particular character is passed in?

Acidity:
All of the character EVE calls require the character_id, but I suppose we could add that as well.

jboning:
Oh, I didn't realize. That's probably good enough then.

jboning:
Hm--but we probably shouldn't 500 if they don't pass it in. Probably success=False and a reason/message. (Or choose an arbitrary key, proxy the request through to the eve API on it, and return the resulting error, heh.)

PR #380, opened 2014-12-08
Add group ACL rule for membership in another group

jboning:
This will make it possible to configure different API key requirements for different alliances.

### DIFF
--- a/brave/core/group/controller.py
+++ b/brave/core/group/controller.py
@@ -133,7 +133,8 @@ class OneGroupController(Controller):
                 g = Group.objects(id=r['group']).first()
                 if not g:
                     # TODO: change dry-run to return a success/fail format so the UI here is better
-                    return 'json:', "ERROR: Couldn't find group {}!!".format(r['group'])
+                    return 'json:', dict(success=False,
+                                         message="ERROR: Couldn't find group {}!!".format(r['group']))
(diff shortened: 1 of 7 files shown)


brave/core/group/controller.py
```diff
@@ -129,20 +129,33 @@ def listify(rule, field):
...
-        log.debug("really!")
         if rule_set == "request":
             self.group.request_rules = rule_objects
         elif rule_set == "join":
             self.group.join_rules = rule_objects
         else:
             self.group.rules = rule_objects
+
+        try:
+            self.group.recursion_check()
```
Acidity: I believe this should be self.group.cycle_check()

Acidity:
When attempting to make a cyclic reference, the ACL check from the server just displays [object Object] rather than an error code.

Acidity:
Another thought, since this only lets you specify one group in the ACL, if you want to do a DENY OR or a GRANT AND on multiple groups you have to make a new group just for that, which isn't great. Not sure how easy it would be to make the ACL support multiple groups though.

Acidity:
Also: deleting a group that another group refers to without deleting the ACL rule first breaks pretty much everything.

jboning:
Note to self, don't merge yet, this can still result in dangling references if you change the ID of a referenced group.

PR #383, opened 2014-12-12
Modify Group requests and join storage to use strings

Acidity:
Joinable and requestable groups had significantly longer execution
times for Group.evaluate prior to this change, as we were
dereferencing the entire join_members and request_members lists from
the database when all we needed to do was check if the user was in it.
We switch to storing only the character names, which are guaranteed to
be unique, and which doesn't require expensive dereferencing while
still meeting all of our intended use cases (being able to easily check
if a character is in the list).


jboning:
Seems sensible! I'm going on faith that you've added .name everywhere that needs it :)

Can you add a note in the upgrading instructions about the migration? Then I will merge

PR #407, opened 2016-04-11
Fix inconsistency in forms styling

bargulg:


bargulg:
Version constraints on webcore package are required because the latest version is incompatible.
We should decide on some reasonable version constraints on all packages in requirements.txt, but this one package is broken right now, so it's pretty urgent

PR #416, opened 2016-05-02
Work around broken key objects

ayust:


bargulg:
even if this is just a workaround, we should at least log which key is broken and which account it belongs to

ayust:
sure, though I suspect it'd be easier to just create a job to check for them every so often rather than spamming logs for a while.

